PR #2, opened 2022-05-12
Implements /records endpoint

tokee:
This implements the `/records` endpoint, making it possible to export batches of records with a given view. The primary use case it to export [JSON Formatted Index Updates](https://solr.apache.org/guide/8_8/uploading-data-with-index-handlers.html#json-formatted-index-updates), making it trivial to populate a running Solr.

The `/records` also allows for mass exports of `JSON-LD` and raw, but those are simple concatenations of the same representations than `/record/{id}` gives. `/records` is special for Solr as it requires qualified output (`add` / `delete`), depending on record state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,3 @@ README_OLD.md
 conf/ds-present-environment.yaml
 conf/ds-present-local.yaml
 setup_project.sh
-/conf/ds-present-kb-collections.yaml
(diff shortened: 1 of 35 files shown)


thomasegense:
I just skimmed all the classes. Too many branches, when all merged I will try understand the unittests.

PR #5, opened 2022-09-02
Wok 398 move iiif presentation to ds-present

VictorHarbo:
This pull request moves IIIF Presentation from ds-image to ds-present.


tokee (commented):
Not at all sure about this: `canvas`, `range`, `annotation` and `list` seems to be part of the `manifest`. Why have explicit endpoints for those, instead of only having the `collection` and `manifest` endpoint. Could you please confer with the resident expert on IIIF on this?

If we want to limit the size of the response, we could add an option specifying which sections the caller is interested in.

VictorHarbo:
Looking at older comments I indeed think you are correct @tokee  `canvas`, `range`, `annotation` and `list` are parts of `manifests`. I will remove these as individual endpoints.

src/main/openapi/openapi_v1.yaml
```diff
@@ -227,10 +227,660 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/Collection'
+  
+# IIIF Presentation endpoints
```
tokee, line 231: Could you add a few words about IIIF Presentation (with links to the standard, which version we support, how complete we are)? Maybe at the top of the OpenAPI YAML

src/main/openapi/openapi_v1.yaml
```diff
@@ -227,10 +227,660 @@ paths:
...
+  /IIIF/{identifier}/collection/{name}:
```
tokee, line 232: This looks... wrong? Is the [collections](https://iiif.io/api/presentation/2.0/#collections) (plural)? And if so, I don't think there should be an `identifier`?

src/main/openapi/openapi_v1.yaml
```diff
@@ -227,10 +227,660 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/collection'
+
+  # TODO: Add option to choose parts of manifest to return.
+  /IIIF/{identifier}/manifest:
+    get:
+      tags:
+        - 'IIIFPresentation'
+      summary: "IIIF Presentation manifest"
```
tokee, line 266: A description whould be nice here. What _is_ a `manifest`?

tokee (changes requested):
Possibly a problem with `collection` and in need of a bit more documentation. That's all.

src/main/openapi/openapi_v1.yaml
VictorHarbo: Correct! It is indeed plural and it does not contain an `identifier`

src/main/openapi/openapi_v1.yaml
VictorHarbo: IIIF Presentation is now described in the top most description.

tokee: The IIIF description is fine in itself, but the API is about more than "just" IIIF. Maybe mention the three formats under `/record/{id}` at the beginning of the presentation and make a second paragraph convering the IIIF-part?

...Okay, not the best suggestion. My point is primarily to avoid the impression that `ds-present` is only about IIIF.

tokee (changes requested):
Just a bit more text, please (see comment in review).

tokee (approved):

PR #6, opened 2022-12-08
Extend the transformer chain with metadata for the record

tokee:
This pull request adds metadata to the chain of DSTransformer`s and is connected to the [extract_copyright_rules](https://github.com/kb-dk/ds-present/tree/extract_copyright_rules) branch.

With thus update, the copyright extractor becomes a `DSTransformer` like the rest. The copyright extractor should

* Derive the needed copyright values from the given xml
* Store the values in the provided `metadata` map
* Return the given xml unchanged

Subsequent `DSTransformer`, such as the one responsible for creating SolrJSON documents, can then use the copyright key-value pairs in the `metadata` map when creating the SolrJSON document.

The setup for such a chain would be something like
```
    - SolrJSON:
        mime: 'application/json'
        transformers:
          - billedsamlingrights:
          - xslt:
              stylesheet: 'xslt/mods2solr.xsl'
```
where the `billedsamlingrights` is the ID for the copyright extractor implementation.

I checked the `extract_copyright_rules` branch and the class `XsltCopyrightMapper` does exactly what is needed. It is only a matter of adjusting how it is called by making that class a `DSTransformer` and adjusting the method signature a bit. I'll do that when this pull request has passed.

### DIFF
--- a/src/main/java/dk/kb/present/DSCollection.java
+++ b/src/main/java/dk/kb/present/DSCollection.java
@@ -109,7 +109,7 @@ public class DSCollection {
     public String getRecord(String recordID, String format) throws ServiceException {
         View view = getView(format);
         String record = storage.getRecord(recordID);
-        return view.apply(record);
+        return view.apply(recordID, record);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


thomasegense (commented):
OK. Still branch does not build, but this is due to the jenkins error.

PR #12, opened 2023-02-06
Scripts for downloading, installing and running a Solr 9 Cloud

tokee:
Scripts ported and upgraded from Solr 8 to Solr 9. README contains instructions for use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,8 @@ After this, a record for ID `doms.radio:albert-einstein.xml` should be delivered
 
 ### Setup Solr
 
-Checkout [ds-solr](https://github.com/kb-dk/ds-solr/), change to the `solr-nested-indexing` branch and follow
-the README, using the Solr setup from `nested-template/conf/`. This boils down to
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -47,11 +47,29 @@ the README, using the Solr setup from `nested-template/conf/`. This boils down t
 ```shell
```
VictorHarbo, line 47: The branch is named `nested-solr-indexing'. Also I misread the readme and thought that I had to start solr from [ds-solr](https://github.com/kb-dk/ds-solr/). However, thats not the case. Maybe specify that a more thorough documentation can be found there, but these three lines should do the trick in ds-present?

VictorHarbo (commented):
The scripts work. However I got a little confused by the documentation as it refers to ds-solr and checking out a branch from that project. I read it as i had to be located in ds-solr to call the scripts mentioned in the readme here. 

Scripts worked flawless when run from the correct place.

README.md
tokee: Whoops, leftover from old instructions.

VictorHarbo (approved):
Looks good to me!

PR #19, opened 2023-03-24
Generalize parameter injection in XSLTTransformer

tokee:
Loading and parsing configuration YAMLs from XSLT is impressive magic but only works when the location and the names of the config files is known. This pull request expands the generic `XSLTTransformer` to handle fixed injections of variables from the config.


VictorHarbo (approved):
Way more general and apply-able than the XSLT hack that I produced! LGTM!

PR #21, opened 2023-03-27
Rename resource_id and make multivalued

VictorHarbo:
Updates name of resource_id in XSLT as discussed earlier today.

### DIFF
--- a/src/main/resources/xslt/mods2solr.xsl
+++ b/src/main/resources/xslt/mods2solr.xsl
@@ -540,12 +540,16 @@
         </xsl:if>
         <!-- Extract image resource-->
         <xsl:if test="m:relatedItem[@type='otherFormat']/m:identifier[@displayLabel='image'][@type='uri']">
-          <xsl:variable name="noPrefix">
-            <xsl:value-of select="f:substring-after(m:relatedItem[@type='otherFormat']/m:identifier[@displayLabel='image'][@type='uri'], 'http://kb-images.kb.dk')"/>
-          </xsl:variable>
-          <f:string key="image_resource">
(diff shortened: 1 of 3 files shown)


src/main/resources/xslt/mods2solr.xsl
```diff
@@ -540,12 +540,16 @@
...
-          <xsl:variable name="noPrefix">
-            <xsl:value-of select="f:substring-after(m:relatedItem[@type='otherFormat']/m:identifier[@displayLabel='image'][@type='uri'], 'http://kb-images.kb.dk')"/>
-          </xsl:variable>
-          <f:string key="image_resource">
-            <xsl:value-of select="substring-before($noPrefix, '.jp')"/>
-          </f:string>
+          <f:array key="resource_id">
+            <xsl:for-each select="m:relatedItem[@type='otherFormat']/m:identifier[@displayLabel='image'][@type='uri']">
+              <xsl:variable name="noPrefix">
+                <xsl:value-of select="f:substring-after(., 'http://kb-images.kb.dk')"/>
```
tokee, line 546: It seems problematic to hardcode the image server URL, demonstrated by the current value being `http` instead of `https`. Could this be adjusted to use a regexp? `https?://[^/]*` should match the server itself.

tokee (changes requested):
The XSLT is a bit too fragile for my taste.

src/main/resources/xslt/mods2solr.xsl
tokee: A kb.kd-internal JIRA has been created for this issue as it is non-critical and as we need the functionality in this pull request for ds-license to work.

PR #23, opened 2023-03-30
Wok 747 update resource id extraction

VictorHarbo:
Updates resource id extraction to use regex and remove weird parts of URL by hand.


tokee (approved):
lgtm

PR #24, opened 2023-07-18
Update solr schema

VictorHarbo:
Changes: 

- Field names have changed quite a lot in solr
- Some new fields have been added
- Small changes in XSLT behavior
- A lot of documentation


thomasegense (commented):
LGTM. 
But fix for timestamp in the category field should be reported as data error to meta-data team.

PR #27, opened 2023-08-21
ID injection for schema.org and Solrdocuments

tokee:
This pull requests sets the external ID for schema.org and Solrdocument representations of DS-resources to the DS-ID rather than the ID from the originating source.

This closes internal issues DISC-46 & 47.


src/test/java/dk/kb/present/solr/EmbeddedSolrTest.java
```diff
@@ -100,7 +100,8 @@ void testRecord000332() throws Exception {
         assertEquals(1, getNumberOfTotalDocuments());
 
         //Full life cycle test
-        SolrDocument record = getRecordById("05fea810-7181-11e0-82d7-002185371280");
+        //SolrDocument record = getRecordById("05fea810-7181-11e0-82d7-002185371280");
```
VictorHarbo, line 103: Maybe we should delete these old getRecordById() calls throughout the class?

VictorHarbo (changes requested):

VictorHarbo (approved):
LGTM!

PR #32, opened 2023-08-24
Add regexp based replace-transformer

tokee:
Makes it possible to perform regexp based adjustment of incoming content.

### DIFF
--- a/src/main/java/dk/kb/present/transform/ReplaceTransformer.java
+++ b/src/main/java/dk/kb/present/transform/ReplaceTransformer.java
@@ -1,0 +1,72 @@
+/*
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


src/main/java/dk/kb/present/transform/ReplaceTransformer.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public String getID() {
+        return ID;
+    }
+
+    @Override
+    public String apply(String s, Map<String, String> metadata) {
+        Matcher m = pattern.matcher(s);
+        return replaceAll ? m.replaceAll(replacement) : m.replaceFirst(replacement);
+    }
```
VictorHarbo, line 64: It is not clear to me why the apply function takes the metadata map, maybe you can explain that to me during lunch?

VictorHarbo (approved):
Looks good. I just have a single perceptual question, which i would like an answer for. 
Approving here

PR #33, opened 2023-08-24
Refactor unit tests to check for existing test files

tokee:
Not sure if this really makes it easier...


VictorHarbo (changes requested):
As discussed, this seems overengineered. Please split into two methods

VictorHarbo (approved):
LGTM

PR #35, opened 2023-08-24
 Add the collection 'radiotv' to ds-present

tokee:
This connects the new `radiotv` collection to the backing `ds-storage` collection `pvica.devel` and connects JSON-LD to `preservica2schemaorg.xsl` & SolrJSON to `preservica2solr.xsl`.

 A temporary replace-hack is used to inject a fake namespace for the faulty Preservica XML blocks. This should be removed when `ds-datahandler` delivers proper XML, either by fixing on the fly or by upstream fixing.

 The IDs for Preservica records does not conform to the ID standard for the DS project. This means that the `/record` endpoint cannot be used. For integration testing the `/records` endpoint can be used.

### DIFF
--- a/conf/ds-present-kb-collections.yaml
+++ b/conf/ds-present-kb-collections.yaml
@@ -69,8 +69,8 @@ config:
 
         # Temporary hack to make the Preservica XML block well-formed (it is missing a namespace declaration)
         testdatahack: &TEST_DATA_HACK
-          regexp: '<xip:DeliverableUnit +status="([^"]+)" *>'
-          replacement: '<xip:DeliverableUnit status="$1" xmlns:xip="http://example.com/">'
+          regexp: '<xip:(DeliverableUnit|Collection|Manifestation) +status="([^"]+)" *>'
+          replacement: '<xip:$1 status="$2" xmlns:xip="http://example.com/">'
           replaceall: false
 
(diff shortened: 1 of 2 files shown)


conf/ds-present-kb-collections.yaml
```diff
@@ -65,6 +65,42 @@ config:
         prefix: 'ds.samlingsbilleder'
         base: 'ds.samlingsbilleder'
         views: *MODS_IMAGE_VIEWS
+    - radiotv:
+
+        # Temporary hack to make the Preservica XML block well-formed (it is missing a namespace declaration)
+        testdatahack: &TEST_DATA_HACK
+          regexp: '<xip:DeliverableUnit +status="([^"]+)" *>'
+          replacement: '<xip:DeliverableUnit status="$1" xmlns:xip="http://example.com/">'
```
VictorHarbo, line 73: I've discovered that the metadata files are not only xip:DeliverableUnit. They can also be xip:Collection and xip:Manifestation. Can you make the regex catch these as well?

VictorHarbo (changes requested):
One minor regex extension

VictorHarbo (approved):
LGTM!

Currently the solr transformations fails on the devel-server, however this is due to my transformation and is not related to this review.

PR #36, opened 2023-08-25
Disc 76 update solr schema

VictorHarbo:
Update solr schema to contain current fields from presevica transformation. 
Please note that some of the license fields from picture metadata had required="true". These have been changed, because they are not present in preservica metadata as it stands for now.

### DIFF
--- a/src/test/resources/solr/dssolr/conf/schema.xml
+++ b/src/test/resources/solr/dssolr/conf/schema.xml
@@ -90,7 +90,8 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?example     Aktuelt?>
   </field>
 
-  <field name="genre" type="string" docValues="true">
+  <!-- TODO: Extract main genre from preservica to this field and make single valued. Collect other genres in a new multivalued field-->
+  <field name="genre" type="string" docValues="true" multiValued="true">
     <?description The genre of the resource?>
     <?example     For a photograph: Portrait?>
     <?example     For a book: novel?>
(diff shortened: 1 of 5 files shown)


src/test/resources/solr/dssolr/conf/schema.xml
```diff
@@ -403,6 +404,44 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?example     Children?>
   </field>
 
+  <!-- Fields that only apply to tv metadata -->
+
+  <field name="original_title" type="string" indexed="true" stored="true">
```
tokee, line 409: `string` is indexed and stored by default. This (and multiple similar fields) could just be `<field name="original_title" type="string">`.

...And we really need to talk about our schema in general, but that is a matter for another session.

src/test/resources/solr/dssolr/conf/schema.xml
```diff
@@ -403,6 +404,44 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
...
+    <?description Identifier from ritzau?>
+    <?example     7f42b813-690e-4897-9553-f6eefe9c4070?>
+  </field>
+
+  <field name="tvmeter_id" type="string" indexed="true">
+    <?description Identifier from tvmeter?>
+    <?example     78e74634-bec1-4cea-a984-20192c97b743?>
+  </field>
+
+  <field name="duration" type="plong" indexed="true" stored="true">
```
tokee, line 424: Maybe `duration_ms` or something else to indicate how this is to be read?

src/test/resources/solr/dssolr/conf/schema.xml
```diff
@@ -90,7 +90,8 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?example     Aktuelt?>
   </field>
 
-  <field name="genre" type="string" docValues="true">
+  <!-- TODO: Extract main genre from preservica to this field and make single valued. Collect other genres in a new multivalued field-->
+  <field name="genre" type="string" docValues="true" multiValued="true">
```
tokee, line 94: The `multiValued="true"` makes `EmbeddedSolrTest.testRecordUldallForTitleAndPlaceOfProductionAndGenre` fail.

tokee (changes requested):
Just a few small things (naming and a failing unit test)

PR #31, opened 2023-08-24
Disc 8 transformation for video

VictorHarbo:
First basic transformation of tv metadata

PR #37, opened 2023-08-25
Disc 102 streamline transformations

VictorHarbo:
This PR streamlines the two transformations from preservica metadata. It handles all of the comments raised in #31 and unifies th structure of the two XSLT files.


tokee (approved):
Nice clean up! I'll merge

PR #39, opened 2023-08-30
fix missing slash in resource_id

VictorHarbo:
adds missing slash in beginning of resource_id

### DIFF
--- a/src/main/resources/xslt/mods2solr.xsl
+++ b/src/main/resources/xslt/mods2solr.xsl
@@ -9,6 +9,7 @@
                xmlns:my="urn:my"
                xmlns:premis="http://www.loc.gov/premis/v3"
                xmlns:mix="http://www.loc.gov/mix/v20"
+               xmlns:functx="http://www.functx.com"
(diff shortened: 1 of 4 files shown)


conf/ds-present-kb-collections.yaml
```diff
@@ -90,6 +90,7 @@ config:
                     injections:
                       #- imageserver: ${path:config.imageservers[default=true].url}
                       - streamingserver: "here_should_be_a_wowza_url"
+                      - old_imageserver: ${path:config.imageservers.oldimageserver.url}
```
tokee, line 93: Sorry, but naming it `oldimageserver` in one place and `old_imageserver` in another place earns you another round in the review process!

tokee: Considering this a second time: No, you are right! While the YAML has the convention of concatenating lowercased words, the XSLT uses underscore. So you did the right thing. My apologies!

tokee (approved):
lgtm

PR #40, opened 2023-08-30
Fix failing resolve of oldimageserver url

tokee:
The old YAML-internal reference was to `${path:config.imageservers.oldimageserver.url}`, but since `imageservers`is a list, it should technically have been `${path:config.imageservers[2].oldimageserver.url}` as `oldimageserver` was the third imageserver in the list. However that was extremely ugly, so I moved `oldimageserver` one level out.


VictorHarbo (approved):
This works! LGTM

PR #41, opened 2023-08-31
Disc 126 fix transformation errors

VictorHarbo:
Fixes multiple errors. One of the being xsi prefixes without namespace declaration


tokee (approved):
Besides a slight naming issue this looks good.

PR #42, opened 2023-08-31
Make IIIF Manifest handle IDs with slash

tokee:
OpenAPI does not handle nonescaped slashed inside of paths (and rightly so as it is against the standard) and the IIIF standard [requires encoding](https://iiif.io/api/image/3.0/#3-identifier). Nevertheless it is accepted practice in the wild to not escape such IDs.

The OpenAPI 3 generator produces a method with this annotation:
```
    @Path("/IIIF/{identifier}/manifest")
```
and thus requires `identifier` to be escaped.

This pull request works around thisby creating a parallel endpoint with manual generated Apache CXF annotation instead of the OpenAPI generated one. CXF has no problem with wildcards in the `Path`, so by annotating a new endpoint with
```
    @Path("/IIIF/{nonescaped:.*}/manifest")
```
the `nonescaped` parameter can contain both escaped and nonescaped slashes.

The version is deployed on the devel server. The manifest returned is just a sample, so asking for any random ID (with or without slashed with or without encoding) should work.

Two things that does not work:

1. jetty does not like _leading_ slashes in the IDs. Or rather, it does not like consecutive slashes `http://example.com/foo//bar`. It is possible to make jetty accept it (See `AMBIGUOUS_EMPTY_SEGMENT` on https://stackoverflow.com/questions/74395500/jetty-returns-400-when-uri-has-2-slashes-in-a-row-while-tomcat-doesnt-care) but I have not discovered how to do so with the Maven Jatty Plugin.
2. Single-escaping `foo%2Fbar` still does not work and produces `Invalid URI: noSlash` in Tomcat (works fine with Jetty). We might be able to to convince our devel-Tomcat to accept it, but I have not explored that.

### DIFF
--- a/src/main/java/dk/kb/present/api/v1/impl/IiifPresentationApiServiceImpl.java
+++ b/src/main/java/dk/kb/present/api/v1/impl/IiifPresentationApiServiceImpl.java
@@ -153,8 +153,7 @@ public class IiifPresentationApiServiceImpl extends ImplBase implements IiifPres
      */
     @Override
     public ManifestDto getPresentationManifest(String identifier) throws ServiceException {
-        // Note the replace that handles double encoding (%252F) of '/' being single-decoded to '%2F'
-        return rawGetPresentationManifest(identifier.replace("%2F", "/"));
+        return rawGetPresentationManifest(identifier);
     }
 
     /*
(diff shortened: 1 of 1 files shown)


VictorHarbo (approved):
lgtm

PR #45, opened 2023-09-01
Disc 130 add origin field

VictorHarbo:


src/main/java/dk/kb/present/View.java
```diff
@@ -40,21 +40,24 @@ public class View extends ArrayList<DSTransformer> implements BiFunction<String,
...
+    private final String origin;
     private final MediaType mime;
 
     /**
      * Creates a view from the given YAML. Expects the YAML to contain a single entry,
      * where the key is the ID for the view and the value is the configuration of the view.
      * @param conf the configuration for this specific view.
      */
-    public View(YAML conf) {
+    public View(YAML conf, String originFromCollection) {
```
tokee, line 51: Remember JavaDoc

src/main/resources/xslt/mods2solr.xsl
```diff
@@ -30,6 +30,7 @@
        <xsl:param name="access_ophavsret_tekst"/>
        <xsl:param name="imageserver"/>
        <xsl:param name="old_imageserver"/>
+       <xsl:param name="origin"/>
```
tokee, line 33: What about the two schema.org XSLTs?

src/test/java/dk/kb/present/TestUtil.java
```diff
@@ -64,15 +69,37 @@ public static String getTransformedFromConfigWithAccessFields(YAML config, Strin
...
+						"injections:\n" +
+						"  - imageserver: 'https://example.com/imageserver/'\n" +
+						"  - old_imageserver: 'http://kb-images.kb.dk'\n" +
+						"  - origin: 'ds.test'\n";
+		YAML yaml = YAML.parse(new ByteArrayInputStream(yamlStr.getBytes(StandardCharsets.UTF_8)));
+		String solrString = getTransformedFromConfigWithAccessFields(yaml, record);
+		Gson gson = new GsonBuilder().setPrettyPrinting().create();
+		JsonElement je = JsonParser.parseString(solrString);
+		String prettyJsonString = gson.toJson(je);
+		System.out.println(prettyJsonString);
```
tokee, line 92: Same code as `prettyPrintSolrJsonFromPreservica`, right? Time for a helper method :-)

tokee (changes requested):
A few small things are noted in the comments, but the big thing missing is a search for `recordBase` in the full code and a subsequent change to `origin`.

Pro tip: Ide IDEA to refactor the `Storage.java` and similar base classes or interfaces: It will suggest a similar rename of the arguments in inheriting classes.

src/main/resources/xslt/mods2solr.xsl
VictorHarbo: Yes, I'll add them as identifiers

tokee (changes requested):
Looks good so far, but still needs a "Find in Files" for `recordBase`.

tokee (changes requested):
That was fast!

`DSStorage` still has some leftovers for `base` and `DSStorageFactory` also used base in the setup, so changes to that one requires updates to the YAML (remember to check is `aegis` also needs an update).

PR #49, opened 2023-09-18
Add warnings when internal test files are absent.

VictorHarbo:


src/test/java/dk/kb/present/transform/XSLTPreservicaToSolrTransformerTest.java
```diff
@@ -28,6 +32,9 @@ String getXSLT() {
 
     @BeforeAll
     public static void beforeMethod() {
+        if (Resolver.getPathFromClasspath("internal_test_files/tvMetadata") == null){
+            log.warn("Internal test files are not present. Unittest 'XSLTPreservicaToSolrTransformerTest' is therefore not run.");
+        }
         Assumptions.assumeTrue(Resolver.getPathFromClasspath("internal_test_files/tvMetadata") != null);
```
tokee, line 38: Alternatively just
```
        Assumptions.assumeTrue(Resolver.getPathFromClasspath("internal_test_files/tvMetadata") != null,
                "Internal test files are not present. Unittest 'XSLTPreservicaSchemaOrgTransformerTest' is therefore not run.");
```
For unknown reasons IntelliJ IDEA eats the `INFO`-level log entry from this but it shows when running `mvn package` from the command line.

tokee (approved):
Low-grade suggestion for code simplification, but nothing blocking.

And nice find with the assumptions. I was not aware of that functionality!

PR #50, opened 2023-09-22
Disc 200 transform manifestations

VictorHarbo:
Add resource_id to deliverable units, which has been extraced by parsing a child manifestation to extract the fileref from here. 

Includes changes in config files. remember to update aegis if deploying to devel environment. This PR applies this functionality to  the preservica2solr transformation, where we are including the child for a parent record. Internal KB issues are created for applying this logic in preservica2schemaorg transformation as well and for implementing the reverse child --> parent logic, which is needed for other types of material.

### DIFF
--- a/src/main/java/dk/kb/present/DSCollection.java
+++ b/src/main/java/dk/kb/present/DSCollection.java
@@ -28,6 +28,8 @@ import dk.kb.util.yaml.YAML;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -117,8 +119,10 @@ public class DSCollection {
      */
     public String getRecord(String recordID, String format) throws ServiceException {
         View view = getView(format);
-        String record = storage.getRecord(recordID);
-        return view.apply(recordID, record);
+        DsRecordDto record = storage.getDSRecord(recordID);
+        String relation = getRelation(record);
+        String recordData = record.getData();
+        return view.apply(recordID, recordData, relation);
     }
 
     /**
@@ -149,7 +153,8 @@ public class DSCollection {
             return storage.getDSRecords(origin, mTime, maxRecords)
                     .peek(record -> {
                         try {
-                            record.data(view.apply(record.getId(), record.getData()));
+                            // TODO: Make this method work with relations as well
+                            record.data(view.apply(record.getId(), record.getData(), ""));
                         } catch (Exception e) {
                             throw new RuntimeTransformerException(
                                     "Exception transforming record '" + record.getId() + "' to format '" + format + "'");
@@ -161,6 +166,29 @@ public class DSCollection {
                      getId(), mTime, maxRecords, e);
             throw new InternalServiceException(
                     "Internal exception requesting records from collection '" + getId() + "' in format " + format);
+        }
+    }
+
+    /**
+     * Extract related records from storage if present in input record.
+     * @param record     to extract relations for.
+     * @return the string value of the related record for the input record.
+     */
+    private String getRelation(DsRecordDto record) {
+        String parentId = record.getParentId();
+        List<String> childrenIds = record.getChildrenIds();
+
+        if (parentId != null){
+            String parent = "";
+            parent = storage.getRecord(record.getParentId());
+            return parent;
+        } else if (childrenIds != null && !childrenIds.isEmpty()){
+            List<String> children = new ArrayList<>();
+            childrenIds.stream().map(id -> children.add(storage.getRecord(id)));
+            // TODO: Make method return correctly for children
+            return children.toString();
+        } else {
+            return "";
         }
     }
 
--- a/src/main/java/dk/kb/present/DSCollection.java
+++ b/src/main/java/dk/kb/present/DSCollection.java
@@ -187,6 +187,9 @@ public class DSCollection {
      */
     private String getChildRecord(DsRecordDto record) {
 
+        // TODO: Figure how to choose correct manifestation for record, if more than one is present
+        // Return first child record, but if there are multiple presentation manifestations,
+        // the rest are currently not added to the transformation
         return record.getChildrenIds() == null ? "" :
                 record.getChildrenIds().stream()
                         .map(id -> getRawUri(id, getRecordEndpoint))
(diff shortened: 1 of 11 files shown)


src/main/java/dk/kb/present/DSCollection.java
```diff
@@ -82,6 +87,8 @@ public class DSCollection {
      */
     private final Map<String, View> views; // keys are lowercase
 
+    private static String getRecordEndpoint;
```
tokee, line 90: Pretty sure you mean `final` not `static`.

src/main/java/dk/kb/present/DSCollection.java
```diff
@@ -97,6 +104,7 @@ public DSCollection(YAML conf, StorageHandler storageHandler) {
             prefix = conf.getString(PREFIX_KEY);
             description = conf.getString(DESCRIPTION_KEY, null);
             storage = storageHandler.getStorage(conf.getString(STORAGE_KEY, null)); // null means default storage
+            getRecordEndpoint = conf.getString(GET_RECORD_ENDPOINT_KEY);
```
tokee, line 107: This makes the get-record-endpoint mandatory for all collections, but we're not using it for anything for most of them (all the picture stuff). How about making it optional by stating `null` as a default and check for it where used?

src/main/java/dk/kb/present/DSCollection.java
```diff
@@ -117,8 +125,10 @@ public DSCollection(YAML conf, StorageHandler storageHandler) {
      */
     public String getRecord(String recordID, String format) throws ServiceException {
         View view = getView(format);
-        String record = storage.getRecord(recordID);
-        return view.apply(recordID, record);
+        DsRecordDto record = storage.getDSRecord(recordID);
+        String relation = getChildRecord(record);
```
tokee, line 129: Maybe `child` instead of `relation`, in anticipation of a future `parent`?

src/main/java/dk/kb/present/DSCollection.java
```diff
@@ -164,6 +175,38 @@ public Stream<DsRecordDto> getDSRecords(Long mTime, Long maxRecords, String form
         }
     }
 
+    /**
+     * Extract children records from storage if present in input record.
+     * @param record     to extract children for.
+     * @return the string value of the related record for the input record.
```
tokee, line 181: Please state that only the first record is handled and that the URI to the raw record is returned.

src/main/java/dk/kb/present/DSCollection.java
```diff
@@ -164,6 +175,38 @@ public Stream<DsRecordDto> getDSRecords(Long mTime, Long maxRecords, String form
...
+
+        if (childrenIds != null && !childrenIds.isEmpty()){
+            childrenIds.forEach(id -> children.add(getRawUri(id)));
+            // TODO: Figure how to choose correct manifestation for record, if more than one is present
+            // Return first child record, but if there are multiple presentation manifestations,
+            // the rest are currently not added to the transformation
+            return children.get(0);
+        } else {
+            return "";
+        }
```
tokee, line 196: Or you could do

```
        return record.getChildrenIds() == null ? "" :
                record.getChildrenIds().stream()
                        .map(DSCollection::getRawUri)
                        .findFirst().orElse("");
```

tokee: (that might be me over complicating things a bit)

src/main/java/dk/kb/present/View.java
```diff
@@ -33,7 +34,7 @@
  * A view is at the core a list of {@link dk.kb.present.transform.DSTransformer}s.
  * It takes a pair of {@code recordID, recordContent} and return transformed recordContent.
```
tokee, line 35: Remember to update the JavaDoc!

src/main/java/dk/kb/present/View.java
```diff
@@ -84,10 +85,11 @@ public MediaType getMime() {
     }
 
     @Override
-    public String apply(String recordID, String content) {
+    public String apply(String recordID, String content, String relation) {
```
tokee, line 88: Unless you means for the `relation` to be a placeholder for either `child` or `parent`, please name the variable more precisely.

src/main/resources/xslt/preservica2solr.xsl
```diff
@@ -35,6 +36,24 @@
...
           <xsl:call-template name="pbc-metadata"/>
         </xsl:for-each>
+
+
+
+        <!-- Manifestations are extracted here. I would like to create a template for this.
+             However, this is quite tricky when using the document() function -->
+        <xsl:if test="$relation != ''">
+          <f:array key="resource_id">
+            <xsl:for-each select="document($relation)/xip:Manifestation/ComponentManifestation/FileRef">
```
tokee, line 46: That was surprisingly direct and usable, seen from the perspective of general XSLT capabilities!

src/test/resources/solr/dssolr/conf/schema.xml
```diff
@@ -445,6 +445,11 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?example     16:9?>
   </field>
 
+  <field name="manifestation_type" type="string">
+    <?description the type of manifestation?>
```
tokee, line 449: You're cheating with this description :-)

**What** is a manifestation?

src/test/resources/test_setup.yaml
```diff
@@ -56,6 +56,7 @@ config:
         prefix: 'local' # The prefix to match on incoming IDs
         origin: 'ds.test'
         description: 'Local records, mostly biographical'
+        getrecordendpoint: 'http://example.com/ds-present/v1/record/'
         # Temporary hack to make the Preservica XML block well-formed (it is missing a namespace declaration)
         testdatahack: &TEST_DATA_HACK_XIP
           regexp: '<xip:(DeliverableUnit|Collection|Manifestation) +status="([^"]+)" *>'
```
tokee, line 62: Is this still needed after the changes to `ds-datahandler`?

tokee (changes requested):
Looks quite good to me. A few suggestions to code, JavaDoc and naming. Only real gripe is with making `getrecordendpoint` mandatory.

src/test/resources/test_setup.yaml
VictorHarbo: Nope, but you are to far ahead. KB internal issue DISC-142 is about removing this. I can remove it it in this PR if you like?

tokee: Sorry, I did not see that it would be handled elsewhere. My bad. But fine to remove it here anyway.

tokee (approved):

tokee (changes requested):
Several unit tests does not run. Most of the because a missing duration?

tokee (approved):

PR #55, opened 2023-10-04
Disc 248 fix genre transformation

VictorHarbo:
Yet another XSLT PR

### DIFF
--- a/src/test/java/dk/kb/present/transform/XSLTPreservicaToSolrTransformerTest.java
+++ b/src/test/java/dk/kb/present/transform/XSLTPreservicaToSolrTransformerTest.java
@@ -65,18 +65,18 @@ public class XSLTPreservicaToSolrTransformerTest extends XSLTTransformerTestBase
 
     @Test
     public void testGenrePresent() {
-        assertContains(RECORD_44979f67, "\"genre\":[\"");
+        assertContains(RECORD_44979f67, "\"genre\":\"");
     }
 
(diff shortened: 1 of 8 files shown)


src/test/java/dk/kb/present/transform/XSLTPreservicaToSolrTransformerTest.java
```diff
@@ -65,7 +65,23 @@ public void testCalculatedDuration() {
...
     @Test
     public void testGenrePresent() {
-        assertContains(RECORD_44979f67, "\"genre\":[\"");
+        assertContains(RECORD_44979f67, "\"genre\":\"");
+    }
+
+    @Test
+    void testGenreContent() {
+        assertContains(RECORD_a8aafb121,"\"list_of_categories\":[\"hovedgenre: Serier\",\"undergenre: Krimiserie\"");
+        assertNotContains(RECORD_a8aafb121, "\"list_of_categories\":[\"hovedgenre: Serier ritzau\",\"undergenre: Krimiserie ritzau\"");
```
tokee, line 74: The `hovedgenre: Serier` notation (not a big fan of the "implicit datastructure hidden in a plain string BTW, but I see the challenge here) is at odds with the examples in the `list_of_categories` field in the Solr `schema.xml`.

tokee (changes requested):
Very little to complain about here. Just a single point.

PR #57, opened 2023-10-05
Disc 249 pvica accession ref

VictorHarbo:


src/test/resources/solr/dssolr/conf/schema.xml
```diff
@@ -490,6 +490,13 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?example     Comedy series?>
   </field>
 
+  <!-- Internal fields for radio/tv -->
```
tokee, line 493: Internal to whom? Maybe mention that the fields are _primarily_ used internally by the Royal Danish Library, but are not classified and might be of use to others.

src/test/resources/solr/dssolr/conf/schema.xml
```diff
@@ -490,6 +490,13 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
...
+  <!-- All fields for internal use are to be prefixed with internal_ -->
+  <field name="internal_accession_ref" type="string">
```
tokee, line 495: Seems like this field should be multi valued?

tokee (commented):
Tiny suggestions. Feel free to ignore if they are too pedantic.

PR #58, opened 2023-10-11
Disc 251 extract all fields

VictorHarbo:
Extracts A LOT of internal fields in both transformations

### DIFF
--- a/src/main/resources/xslt/preservica2schemaorg.xsl
+++ b/src/main/resources/xslt/preservica2schemaorg.xsl
@@ -10,6 +10,7 @@
                xmlns:padding="http://kuana.kb.dk/types/padding/0/1/#"
                xmlns:access="http://doms.statsbiblioteket.dk/types/access/0/1/#"
                xmlns:pidhandle="http://kuana.kb.dk/types/pidhandle/0/1/#"
+               xmlns:program_structure="http://doms.statsbiblioteket.dk/types/program_structure/0/1/#"
                version="3.0">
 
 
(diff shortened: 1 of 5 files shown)


src/main/resources/xslt/preservica2solr.xsl
```diff
@@ -76,7 +150,6 @@
           </f:string>
         </xsl:for-each>
       </f:array>
-
       <xsl:for-each select="pbcoreGenre/genre">
         <xsl:choose>
           <xsl:when test="f:contains(., 'hovedgenre:')">
```
tokee, line 155: `for-each` indicates that there can be more than one genre, but in the `schema.xml`, `genre` is single-valued? If Preservica guarantees that there is only a single `hovedgenre` and a single `undergenre`, please state this in a comment in the XSLT.

src/main/resources/xslt/preservica2solr.xsl
```diff
@@ -91,16 +164,17 @@
...
+         It's rather a description of the type of resource. -->
     <f:string key="resource_description">
       <xsl:value-of select="pbcoreInstantiation/formatMediaType"/>
     </f:string>
 
-
+    <!-- Title extraction -->
     <xsl:for-each select="pbcoreTitle">
-      <xsl:if test="titleType = 'titel'">
+      <xsl:if test="titleType = 'titel' and title != ''">
```
tokee, line 177: The need for a check for empty string is a bit worrying. Are there other properties that can be defined in the Preservica output that can have empty values?

src/test/resources/solr/dssolr/conf/schema.xml
```diff
@@ -484,20 +505,151 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
...
+    <?example     true?>
+  </field>
+
+  <field name="has_subtitles_for_hearing_impaired" type="boolean">
+    <?description A boolean value for subtitles for hearing impaired. If true, then the resource contains subtitles.?>
+    <?example     true?>
+  </field>
+
+  <field name="pid" type="string">
+    <?description A percistent identifier (PID) is a long lasting reference to the resource.?>
```
tokee, line 524: "persistent" :-)

tokee: And maybe mention that PIDs can be resolved from e.g. http://hdl.handle.net/ ?

tokee (changes requested):
Hard to review without knowing the formats well, but overall it looks good and the `schema.xml` seems to be in sync with the produced Solr Documents - I only tried indexing one document.

A few comments are inlined.

src/test/resources/solr/dssolr/conf/schema.xml
```diff
@@ -595,6 +595,11 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?example     false?>
   </field>
 
+  <field name="internal_showviewcode" type="string">
+    <?description ?> <!-- TODO: what is this?-->
+    <?example     ?> <!-- TODO: what is this?-->
```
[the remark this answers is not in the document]
VictorHarbo: Historical fact of the day!

src/main/resources/xslt/preservica2schemaorg.xsl
```diff
@@ -127,10 +148,14 @@
...
       </xsl:if>
     </xsl:for-each>
-
-    <!-- Saves all extensions in a variable used to check if one or more conditions are met in any of them.
-         This is done to create one nested object in the JSON with values from multiple PBC extensions. -->
-    <xsl:variable name="pbcExtensions" select="./pbcoreExtension/extension"/>
+    
+    
+    <!-- Creates datePublished, when pbcore extension tells that the program is a premiere.  -->
+    <xsl:if test="$pbcExtensions[f:contains(., 'premiere:premiere')]">
```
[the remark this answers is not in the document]
VictorHarbo: Good catch

src/test/java/dk/kb/present/transform/XSLTPreservicaSchemaOrgTransformerTest.java
```diff
@@ -142,6 +167,65 @@ void testAbstractCreation() throws IOException {
...
+                                                "\"kb:retransmission\":\"false\"," +
+                                                "\"kb:maingenre_id\":\"1\"," +
+                                                "\"kb:channel_id\":\"3\"," +
+                                                "\"kb:country_of_origin_id\":\"0\"," +
+                                                "\"kb:ritzau_program_id\":\"25101114\"," +
+                                                "\"kb:program_ophold\":\"false\"," +
+                                                "\"kb:subgenre_id\":\"708\"," +
+                                                "\"kb:episode_id\":\"0\"," +
+                                                "\"kb:season_id\":\"0\"," +
+                                                "\"kb:series_id\":\"0\"," +
```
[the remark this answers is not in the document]
VictorHarbo: I am not sure. They are stated as zeroes in the pvica metadata. Filtering zeros seems wrong as well. I believe that zero currently represents that the resource isn't part of a series or season, but I could be wrong?

src/main/resources/xslt/preservica2solr.xsl
VictorHarbo: I'll add the comment. Just out of curiosity, would you prefer two if statements instead?

src/main/resources/xslt/preservica2solr.xsl
VictorHarbo: I am not sure about this, but my initial thought is that many fields could be generated without content. I believe the fields are created even though they are empty, so that curators can create the metadata in Kuana - I can point you to some internal documentation of this.

src/test/java/dk/kb/present/transform/XSLTPreservicaSchemaOrgTransformerTest.java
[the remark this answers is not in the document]
tokee: Since they are delivered from Preservica, let's keep them.

src/main/resources/xslt/preservica2solr.xsl
tokee: No thank you :-)

This is too much nitpicking in a semi-random direction from my side.

src/main/resources/xslt/preservica2solr.xsl
tokee: No. `choose` is equivalent to a `switch` and that signals a "trivial" check between a finite list of match-values. `if` is open and signals a more flexible approach where secondary considerations are possible. But we're really at the finer points of conditional choosing here.

PR #63, opened 2023-10-16
Disc 258 274 and 314

VictorHarbo:
Combined review of 3 jiras. 
Two about XSLT changes, one about organizing solr schema

### DIFF
--- a/src/main/resources/xslt/preservica2schemaorg.xsl
+++ b/src/main/resources/xslt/preservica2schemaorg.xsl
@@ -90,32 +90,19 @@
       <xsl:value-of select="pbcoreTitle[2]/title"/>
     </xsl:variable>
 
-<!--    TODO: name and alternate name does not require the value map - they are to be specified as strings-->
     <xsl:choose>
       <xsl:when test="$title = $original-title or ($title != '' and $original-title = '')">
-        <f:array key="name">
-          <f:map>
-            <f:string key="value">
-              <xsl:value-of select="$title"/>
-            </f:string>
-          </f:map>
-        </f:array>
+        <f:string key="name">
+          <xsl:value-of select="$title"/>
+        </f:string>
       </xsl:when>
       <xsl:otherwise>
-        <f:array key="name">
-          <f:map>
-            <f:string key="value">
-              <xsl:value-of select="$title"/>
-            </f:string>
-          </f:map>
-        </f:array>
-        <f:array key="alternateName">
-          <f:map>
-            <f:string key="value">
-              <xsl:value-of select="$original-title"/>
-            </f:string>
-          </f:map>
-        </f:array>
+        <f:string key="name">
+          <xsl:value-of select="$title"/>
+        </f:string>
+        <f:string key="alternateName">
+          <xsl:value-of select="$original-title"/>
+        </f:string>
       </xsl:otherwise>
     </xsl:choose>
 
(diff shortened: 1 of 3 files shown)


src/main/resources/xslt/preservica2schemaorg.xsl
```diff
@@ -90,32 +90,19 @@
...
-        <f:array key="name">
-          <f:map>
-            <f:string key="value">
-              <xsl:value-of select="$title"/>
-            </f:string>
-          </f:map>
-        </f:array>
+        <f:string key="name">
+          <xsl:value-of select="$title"/>
+        </f:string>
```
tokee, line 97: If both `title` and `original-title` are empty, there will be an empty entry  here. Is that intentional?

src/main/resources/xslt/preservica2schemaorg.xsl
```diff
@@ -90,32 +90,19 @@
...
-              <xsl:value-of select="$original-title"/>
-            </f:string>
-          </f:map>
-        </f:array>
+        <f:string key="name">
+          <xsl:value-of select="$title"/>
+        </f:string>
+        <f:string key="alternateName">
+          <xsl:value-of select="$original-title"/>
+        </f:string>
```
tokee, line 105: If `title` is empty and `original-title` is not, there will be an empty `name` entry here. Is that intentional?

src/main/resources/xslt/preservica2schemaorg.xsl
```diff
@@ -375,32 +362,32 @@
...
         <xsl:when test="/xip:DeliverableUnit/Metadata/pbc:PBCoreDescriptionDocument/pbcoreInstantiation/formatChannelConfiguration = 'ikke surround'">
-          <f:string key="kb:surround_sound"><xsl:value-of select="false()"/></f:string>
+          <f:boolean key="kb:surround_sound"><xsl:value-of select="false()"/></f:boolean>
         </xsl:when>
       </xsl:choose>
       <!-- Create boolean for color-->
       <xsl:choose>
         <xsl:when test="/xip:DeliverableUnit/Metadata/pbc:PBCoreDescriptionDocument/pbcoreInstantiation/formatColors = 'farve'">
-          <f:string key="kb:color">true</f:string>
+          <f:boolean key="kb:color">true</f:boolean>
```
tokee, line 374: I don't understand why `<xsl:value-of select="false()"/>` is used in some places and `true` in other places!?

tokee (commented):
Thank you for the clean up of both types and schema. My comment are more questions, so if they are not trivial to answer we should take them over Zoom.

src/main/resources/xslt/preservica2schemaorg.xsl
VictorHarbo: Nope. non-intentional, I will add a check for empty strings

src/main/resources/xslt/preservica2schemaorg.xsl
VictorHarbo: Nope. non-intentional, I will add a check for empty strings

src/main/resources/xslt/preservica2schemaorg.xsl
VictorHarbo: Sloppiness from my side. I'll convert to using <xsl:value-of select="bool()"/> everywhere.

PR #64, opened 2023-10-19
Disc 295 call new storage

VictorHarbo:
Change method calling storage and update XSLTs to use child data directly


thomasegense (approved):
LGTM!

getNewestChild needs to be discussed in another issue.

PR #65, opened 2023-10-20
Disc 323 harden xslts

VictorHarbo:
Mostly add checks for empty strings and normalization of spaces. Remember to pull aegis and do kb init before running tests


thomasegense (approved):
XSLT's just accepted without understanding logic.

Can you see also removed some of the too spammy logging from the unittests.

PR #71, opened 2023-10-31
Disc 335 add filtering

VictorHarbo:


tokee:
Marking this as draft as discussion between @VictorHarbo & @tokee shows that the `View.apply`-method should be refactored to take `DsRecordDto` instead of `recordID, recordData, childData`. After refactoring it will be up to the `View.apply` to perform filtering or adjustments on the received record to derive e.g. child records.

src/main/java/dk/kb/present/View.java
```diff
@@ -186,6 +186,7 @@ private String getFirstChild(DsRecordDto record) {
         return record.getChildren() == null ? "" :
                 record.getChildren().stream()
                         .map(this::getNonNullChild)
+                        .filter(childData -> childData.contains("<ManifestationRelRef>2</ManifestationRelRef>"))
```
tokee, line 189: Method name + Javadoc needs to be changed to reflect the new logic.

tokee (changes requested):
This is starting to look very clean & nice. Only one request for change.

PR #72, opened 2023-10-31
Disc 363 refactor view

VictorHarbo:


src/main/java/dk/kb/present/View.java
```diff
@@ -63,6 +65,7 @@ public View(YAML conf, String originOfCollection) {
         conf = conf.getSubMap(id);
         String[] mimeTokens = conf.getString(MIME_KEY).split("/", 2);
         mime = new MediaType(mimeTokens[0], mimeTokens[1]);
+        strategy = conf.getString(STRATEGY_KEY, "none");
```
tokee, line 68: There are a few and clearly defines strategies, so defining them as an `enum` is the best ... strategy. This is also a clean way to describe what the different strategies are about, via Javadoc.

src/main/java/dk/kb/present/View.java
```diff
@@ -85,30 +88,55 @@ public MediaType getMime() {
...
+    public String apply(DsRecordDto record) {
+        final Map<String, String> metadata = createBasicMetadataMap(record);
 
-        if (child != null ) {
-            metadata.put("childRecord", child);
-        }
+        String content = record.getData();
+
+        switch (strategy) {
+            case "manifestation":
```
tokee, line 97: There are no strategies defined in the `ds-present-kb-collections.yaml`. Maybe it could be injected there for a few `View`s and with a comment explaining what it is about, in order to make it easier to create other setups?

src/main/java/dk/kb/present/View.java
```diff
@@ -85,30 +88,55 @@ public MediaType getMime() {
...
+    }
+
+    /**
+     * Update the map of metadata with child record from the deliverable unit contained in the input record.
+     * @param record with an expanded tree containing manifestation childs.
+     * @param metadata map that values from the record is extracted to.
+     */
+    private void updateMetadataMapWithPreservicaManifestation(DsRecordDto record, Map<String, String> metadata) {
+        String child = getFirstChild(record);
+        if (child != null ) {
```
tokee, line 135: `child` is never null.

src/main/java/dk/kb/present/View.java
```diff
@@ -118,4 +146,31 @@ public String toString() {
                ", transformers=" + super.toString() +
                ')';
     }
+
+    /**
+     * If record has children, the first child is returned.
```
tokee, line 151: Also note what is returned if there are no children.

src/main/java/dk/kb/present/View.java
```diff
@@ -118,4 +146,31 @@ public String toString() {
...
+        // Return first child record, but if there are multiple presentation manifestations,
+        // the rest are currently not added to the transformation
+        return record.getChildren() == null ? "" :
+                record.getChildren().stream()
+                        .map(this::getNonNullChild)
+                        .findFirst().orElse("");
+    }
+
+    /**
+     * Check for child being null.
```
tokee, line 166: No, it checks for `child.getData()` being `null`.

src/main/java/dk/kb/present/View.java
```diff
@@ -118,4 +146,31 @@ public String toString() {
...
+     * @param record to extract the newest child from.
+     * @return  the data from the first child related to the input record.
+     */
+    private String getFirstChild(DsRecordDto record) {
+        // TODO: Figure how to choose correct manifestation for record, if more than one is present
+        // Return first child record, but if there are multiple presentation manifestations,
+        // the rest are currently not added to the transformation
+        return record.getChildren() == null ? "" :
+                record.getChildren().stream()
+                        .map(this::getNonNullChild)
```
tokee, line 161: Is this the proper logic for Preservica-records? I thought that the type of manifestation played a role!?

src/main/java/dk/kb/present/View.java
```diff
@@ -118,4 +146,31 @@ public String toString() {
                ", transformers=" + super.toString() +
```
tokee, line 146: Please add `strategy` to the `toString`-method.

src/main/java/dk/kb/present/View.java
```diff
@@ -85,30 +88,55 @@ public MediaType getMime() {
...
+
+    /**
+     * Update the map of metadata with child record from the deliverable unit contained in the input record.
+     * @param record with an expanded tree containing manifestation childs.
+     * @param metadata map that values from the record is extracted to.
+     */
+    private void updateMetadataMapWithPreservicaManifestation(DsRecordDto record, Map<String, String> metadata) {
+        String child = getFirstChild(record);
+        if (child != null ) {
+            metadata.put("childRecord", child);
```
tokee, line 136: As we are very specific here, maybe call it `manifestation` instead of `childRecord`?

src/main/java/dk/kb/present/View.java
```diff
@@ -85,30 +88,55 @@ public MediaType getMime() {
...
+     */
+    private Map<String,String> createBasicMetadataMap(DsRecordDto record) {
+        final Map<String, String> metadata = new HashMap<>();
+        metadata.put("recordID", record.getId());
+        metadata.put("origin", origin);
+        return metadata;
+    }
+
+    /**
+     * Update the map of metadata with child record from the deliverable unit contained in the input record.
```
tokee, line 129: "manifestation record" instead of "child record"?

src/main/java/dk/kb/present/View.java
```diff
@@ -85,30 +88,55 @@ public MediaType getMime() {
...
-            metadata.put("childRecord", child);
-        }
+        String content = record.getData();
+
+        switch (strategy) {
+            case "manifestation":
+                updateMetadataMapWithPreservicaManifestation(record, metadata);
+            case "none":
 
+        }
```
tokee, line 101: When using a `switch`, please add a `default` that throws an `UnsupportedOperationException` with the given value.

tokee (changes requested):
Multiple minor stuff. Overall is seems sound.

src/main/java/dk/kb/present/View.java
VictorHarbo: It does play a role! However, I believe that is handled in this PR https://github.com/kb-dk/ds-present/pull/71, which we converted to a draft until this one is through?

src/main/java/dk/kb/present/View.java
tokee: There's quite a lot of _silent knowledge_ in the Javadoc for `MANIFESTATION`. Without turning it into Moby Dick, could you mention that it is oriented towards Preservica and what the DeliverableUnit/Manifestation relationship is?

src/main/java/dk/kb/present/View.java
tokee: Ah! Of course. My bad.

src/main/java/dk/kb/present/View.java
tokee: Nice error message :heart:

PR #73, opened 2023-11-03
Add license support to ds-present

tokee:
This pull request adds support for ds-license integration (enabled as default). The 2 endpoints `/record({id}` and `/records` has different behaviour

* `/record/{id}` throws an exception (not found if the record is not in the Sol index, forbidden if access is not allowed) if the record cannot/must not be delivered
* `/records` passes existing allowed records and silently ignores non-indexed/non-access records

When testing it is possible to turn off access checking in the configuration.

* Note 1: The `DISC-367_license`-branch is deployed on the devel server
* Note 2: `aegis` has been updated

### DIFF
--- a/src/main/java/dk/kb/present/PresentFacade.java
+++ b/src/main/java/dk/kb/present/PresentFacade.java
@@ -14,6 +14,7 @@
  */
 package dk.kb.present;
 
+import dk.kb.present.api.v1.impl.DsPresentApiServiceImpl;
 import dk.kb.present.config.ServiceConfig;
 import dk.kb.present.model.v1.CollectionDto;
 import dk.kb.present.model.v1.ViewDto;
@@ -33,10 +34,10 @@ import org.slf4j.LoggerFactory;
 
 import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.core.StreamingOutput;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Locale;
+import java.util.*;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 
 /**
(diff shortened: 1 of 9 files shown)


src/main/java/dk/kb/present/PresentFacade.java
```diff
@@ -181,7 +206,8 @@ private static StreamingOutput getRecordsFull(
 
     // Direct ds-storage record JSON
     private static StreamingOutput getRecordsSolr(DSCollection collection, Long mTime, Long maxRecords,
-                                                  HttpServletResponse httpServletResponse) {
+                                                  HttpServletResponse httpServletResponse,
+                                                  Function<List<DsRecordDto>, Stream<DsRecordDto>> accessPredicate) {
```
VictorHarbo, line 210: All other methods calls the function accessFilter. Maybe choose one or the other?

src/main/java/dk/kb/present/api/v1/impl/DsPresentApiServiceImpl.java
```diff
@@ -21,6 +32,30 @@
 public class DsPresentApiServiceImpl extends ImplBase implements DsPresentApi {
     private static final Logger log = LoggerFactory.getLogger(DsPresentApiServiceImpl.class);
 
+    // License handling in DsPresentApiServiceImpl as tokens from the request will be needed later on
+    private static final String LICENSE_URL_KEY = "config.licensemodule.url";
+    private static final String LICENSE_ALLOWALL_KEY = "config.licensemodule.allowall";
```
VictorHarbo, line 37: Just a comment: It would be so pretty if we removed the redundant `config.` keys from config files.

src/main/java/dk/kb/present/api/v1/impl/DsPresentApiServiceImpl.java
```diff
@@ -21,6 +32,30 @@
...
+    static DsLicenseApi licenseClient;     // Shared between instances
+    static boolean licenseAllowAll = ServiceConfig.getConfig().getBoolean(LICENSE_ALLOWALL_KEY, false);
+    public static final String RECORD_ACCESS_TYPE = "Search"; // TODO: Evaluate if a specific type is needed
```
VictorHarbo, line 40: Is this TODO relevant for this PR?

src/main/java/dk/kb/present/api/v1/impl/DsPresentApiServiceImpl.java
```diff
@@ -102,10 +152,117 @@ public StreamingOutput getRecords(String collection, Long mTime, Long maxRecords
...
+                    return new ArrayList<>(ids);
+                }
+                Set<String> allowed = new HashSet<>(response.getAccessIds());
+                return ids.stream().filter(allowed::contains).collect(Collectors.toList());
+            }
+            // TODO: Should a warning be logged here? Will getAccessIds return null if no IDs are allowed?
+            return Collections.emptyList();
+        };
+    }
+
```
VictorHarbo, line 267: Would it make sense to move these methods to own class?

VictorHarbo (changes requested):
Minor suggestions. Super nice mockito testing!

src/main/java/dk/kb/present/PresentFacade.java
tokee: Leftover from previous code. Nice catch.

VictorHarbo (approved):
LGTM! Code is neat!

PR #74, opened 2023-11-07
Move collection specific behaviour to config and remove old endpoint

VictorHarbo:


src/main/java/dk/kb/present/DSCollection.java
```diff
@@ -87,7 +87,7 @@ public class DSCollection {
      */
     private final Map<String, View> views; // keys are lowercase
 
-    private String getRecordEndpoint;
+    private RecordTypeDto recordType;
```
tokee, line 90: Isn't this more of a `recordRequestType` instead of just `recordType`? I was confused at first. Maybe add a bit of Javadoc to say what it is used for?

src/main/java/dk/kb/present/DSCollection.java
```diff
@@ -99,17 +99,13 @@ public class DSCollection {
     public DSCollection(YAML conf, StorageHandler storageHandler) {
         id = conf.keySet().stream().findFirst().orElseThrow();
         try {
-            origin = conf.getSubMap(id).getString("origin");
             conf = conf.getSubMap(id); // There must be some properties for a storage
+            origin = conf.getString("origin");
             prefix = conf.getString(PREFIX_KEY);
             description = conf.getString(DESCRIPTION_KEY, null);
+            recordType = RecordTypeDto.valueOf(conf.getString(RECORDTYPE_KEY));
```
tokee, line 106: This makes the retrieval type mandatory. This makes 100% sense for transforming views, but if `StorageRecords-Lines` are requested, wouldn't the caller want all types? Or maybe we should reconsider the `StorageRecords-Lines` option and maybe make it an explicit endpoint for exporting "raw" data?

tokee (commented):
Nice clean up. One minor comment about naming & javadoc and one bigger "We'll have to rethink this". The latter calls for a new JIRA.

src/main/java/dk/kb/present/DSCollection.java
VictorHarbo: You're right!

src/main/java/dk/kb/present/DSCollection.java
VictorHarbo: I like the idea of a raw-endpoint

PR #75, opened 2023-11-08
Disc 302 transform schema to solr

VictorHarbo:
Quite big pr. Please focus on the new schemaog2solr transformation and how it gets used and applied throughout the code.


src/main/java/dk/kb/present/View.java
```diff
@@ -100,6 +102,12 @@ public View(YAML conf, String originOfCollection) {
                 throw new RuntimeException(e); // Wrap for stream use
             }
         }
+
+        try {
+            placeholderXml = Resolver.resolveUTF8String("placeholder.xml");
```
tokee, line 107: As the content of `placeholder.xml` is tiny, maybe just set it directly as a String instead of storing it in a file? That way it also becomes more transparent what it is.

src/main/java/dk/kb/present/View.java
```diff
@@ -128,9 +136,15 @@ public String apply(DsRecordDto record) {
 
         }
 
+
         for (DSTransformer transformer: this) {
             try {
-                content = transformer.apply(content, metadata);
+                if (transformer.getStylesheet() != null && transformer.getStylesheet().contains("schemaorg2solr.xsl")){
+                    metadata.putIfAbsent("schemaorgjson", content);
+                    content = transformer.apply(placeholderXml, metadata);
```
tokee, line 144: I don't understand why it is necessary to use an empty placeholder here. Why can't the original content be used (and ignored in the XSLT)?

There's probably a very fine and a bit complicated reason for this, so we could take that face2face.

src/main/java/dk/kb/present/View.java
```diff
@@ -128,9 +136,15 @@ public String apply(DsRecordDto record) {
 
         }
 
+
         for (DSTransformer transformer: this) {
             try {
-                content = transformer.apply(content, metadata);
+                if (transformer.getStylesheet() != null && transformer.getStylesheet().contains("schemaorg2solr.xsl")){
+                    metadata.putIfAbsent("schemaorgjson", content);
```
tokee, line 143: Nope, sorry, no hacking allowed 👮. The behaviour should be generalized and specified in the properties for the workflow. Alternatively a subclass of `XSLTTransformer` could be created.

src/main/java/dk/kb/present/transform/DSTransformer.java
```diff
@@ -32,4 +32,11 @@ public abstract class DSTransformer implements BiFunction<String, Map<String, St
      */
     abstract public String getID();
 
+    /**
+     * @return the stylesheet used by the transformer, if any existst. If stylesheet does not exist return null.
+     */
+    abstract public String getStylesheet();
```
tokee, line 38: We should make `DSTransformer` an `interface` and specify the `getStylesheet()` as a `default`. That way most of the `DSTransformer`-classes won't have to override it.

tokee (changes requested):
For once there was a real problem in the code (the hacking part). And some XSLT incantations you'll have to explain to me.

src/main/java/dk/kb/present/View.java
VictorHarbo: That's actually a smart idea! Haven't thought about it

VictorHarbo: Turns out I have thought about it. The content needs to be valid XML and not JSON, therefore the placeholder

src/main/java/dk/kb/present/transform/XSLTSolrFromSchemaTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+
+    static {
+        System.setProperty("javax.xml.transform.TransformerFactory", "net.sf.saxon.TransformerFactoryImpl");
+        transformerFactory = TransformerFactoryImpl.newInstance();
+        /* Attributes on the transformer level can be tweaked quite a bit, however the javax
+           TransformerFactory is oriented towards XSLT 1.0, which doesn't allow calling an XSLT
+           without giving it XML as input. The only way to transform through this API is with either an empty placeholder
+           or encapsulating the JSON structure in XML tags.
+           Link to StackOverfow discussion on this: https://stackoverflow.com/a/35383155/12400491 */
+        // transformerFactory.setAttribute("http://saxon.sf.net/feature/initialTemplate", "initial-template");
```
tokee, line 37: Very nice!

src/test/java/dk/kb/present/ViewTest.java
```diff
@@ -62,6 +62,20 @@ void jsonldPvica() throws Exception {
         assertTrue(jsonld.contains("\"name\":\"Før Bjørnen Er Skudt\""));
     }
 
+    @Test
+    void solrFromPvica() throws Exception {
+        YAML conf = YAML.resolveLayeredConfigs("test_setup.yaml");
+        YAML radiotvConf = conf.getYAMLList(".config.collections").get(1);
+        View solrView = new View(radiotvConf.getSubMap("radiotv").getYAMLList("views").get(2), radiotvConf.getSubMap("radiotv").getString("origin"));
+        String pvica = Resolver.resolveUTF8String("internal_test_files/tvMetadata/df3dc9cf-43f6-4a8a-8909-de8b0fb7bd00.xml");
```
tokee, line 70: There's an awful lot of "we know what the setup file contains"-magic here and in the other tests. We should consider a more resilient approach. But not in this pull request.,

tokee (changes requested):
Only thing that does not look good is the organization of unit test methods for Preservica-material.

src/test/java/dk/kb/present/ViewTest.java
VictorHarbo: These are to be known as "YAML incantations"

src/test/java/dk/kb/present/transform/XSLTTransformerTestBase.java
```diff
@@ -122,13 +151,40 @@ public final void assertMultiTests(String record, Consumer<String>... tests) {
...
+        try {
+            solrString = TestUtil.getTransformedToSolrJsonThroughSchemaJson(PRESERVICA2SCHEMAORG, record);
+        } catch (Exception e) {
+            throw new RuntimeException(
+                    "Unable to fetch and transform '" + record + "' using XSLT '" + getXSLT() + "'", e);
+        }
+
+        //TestUtil.prettyPrintJson(solrString);
+
+        Arrays.stream(tests).forEach(test -> test.accept(solrString));
```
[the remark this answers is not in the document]
VictorHarbo: Correct

PR #76, opened 2023-11-09
Remove deprecated ExportWriterFactory

tokee:
The newly rolled kb-util 1.4.15 contains a compatible `ExportWriterFactory`, making it possible to switch to that instead of having duplicate code.

Temporarily lowering to 1.4.14 and running the unit tests will trigger a fail, demonstrating that the relevant part of the `ExportWriterFactory` is indeed tested.


VictorHarbo (approved):
LGTM!

PR #78, opened 2023-11-10
Disc 350 move testfiles

VictorHarbo:
Moves definition of test files from being divided between different test classes to one common class


thomasegense (approved):
accepted

PR #79, opened 2023-11-10
Disc 358 xslt corrections

VictorHarbo:
Removes null values from schema keywords/ solr categories

### DIFF
--- a/src/main/java/dk/kb/present/DSCollection.java
+++ b/src/main/java/dk/kb/present/DSCollection.java
@@ -175,9 +175,9 @@ public class DSCollection {
         log.debug("Calling storage.getDSRecords(origin='{}', mTime={}, maxRecords={})",
                 origin, mTime, maxRecords);
         try {
-            // 500 is a magic number, which is poor code style. Currently, it controls batch size against ds-license
+            // 35 is a magic number, which is poor code style. Currently, it controls batch size against ds-license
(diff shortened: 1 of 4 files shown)


src/main/java/dk/kb/present/DSCollection.java
```diff
@@ -175,9 +175,9 @@ public Stream<DsRecordDto> getDSRecords(
         log.debug("Calling storage.getDSRecords(origin='{}', mTime={}, maxRecords={})",
                 origin, mTime, maxRecords);
         try {
-            // 500 is a magic number, which is poor code style. Currently, it controls batch size against ds-license
+            // 35 is a magic number, which is poor code style. Currently, it controls batch size against ds-license
             return ExtractionUtils.splitToLists(
-                            storage.getDSRecordsByRecordTypeLocalTree(origin, recordRequestType, mTime, maxRecords), 500)
+                            storage.getDSRecordsByRecordTypeLocalTree(origin, recordRequestType, mTime, maxRecords), 35)
```
tokee, line 180: Note to self @VictorHarbo : Increase this when the problem has been fixed in the `ds-license` end.

tokee (approved):
This did require a removal of old test values and `kb init`. ~~We~~ I really need to fix that to happen automatically. Besides that LGTM!

PR #81, opened 2023-11-13
Disc 400 log manifestation warn

VictorHarbo:


src/main/java/dk/kb/present/View.java
```diff
@@ -187,11 +188,28 @@ private String getFirstPresentationManifestation(DsRecordDto record) {
...
+     * @return the first presentation manifestation for a record.
+     */
+    private String returnPresentationManifestationFromList(List<String> presentationManifestations, String recordId) {
+        if (presentationManifestations.size() > 1) {
+            log.warn("Multiple presentation manifestations were present for record with id: '{}'. " +
+                    "Only the first has been returned", recordId);
+        }
+
+        return presentationManifestations.get(0);
     }
```
tokee, line 213: I like this separation as it is also a preparation for later code which might select the manifestation using a more elaborate schema than "Meh. Just pick the first one".

tokee (approved):
Nothing but praise here. I'll merge.

PR #82, opened 2023-11-13
Bugfix: Copy Solr setup files to target for testing

tokee:
The previous setup where `SOLR_HOME` was `src/main/solr/` caused Solr to create a sub-folder named `data` holding the index files. This commit changes `SOLR_HOME` to be `target/solr/`.

### DIFF
--- a/src/test/java/dk/kb/present/solr/EmbeddedSolrFieldAnalyseTest.java
+++ b/src/test/java/dk/kb/present/solr/EmbeddedSolrFieldAnalyseTest.java
@@ -33,7 +33,7 @@ import org.slf4j.LoggerFactory;
 public class EmbeddedSolrFieldAnalyseTest {
 
 	private static final Logger log = LoggerFactory.getLogger(EmbeddedSolrTest.class);
-	private static String solr_home = "src/main/solr";
+	private static String solr_home = "target/solr/";
 
 	private static CoreContainer coreContainer = null;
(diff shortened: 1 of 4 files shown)


src/test/java/dk/kb/present/solr/EmbeddedSolrFieldAnalyseTest.java
```diff
@@ -33,7 +33,7 @@
 public class EmbeddedSolrFieldAnalyseTest {
 
 	private static final Logger log = LoggerFactory.getLogger(EmbeddedSolrTest.class);
-	private static String solr_home = "src/main/solr";
+	private static String solr_home = "target/solr/";
```
VictorHarbo, line 36: here you are defining the path with a trailing slash, and below its done without it. It probably doesn't matter, but should we define them in the same way?

src/test/java/dk/kb/present/solr/EmbeddedSolrTest.java
```diff
@@ -56,7 +56,7 @@
 public class EmbeddedSolrTest {
 
     private static final Logger log = LoggerFactory.getLogger(EmbeddedSolrTest.class);
-    private static final String solr_home = "src/main/solr";
+    private static final String solr_home = "target/solr";
```
VictorHarbo, line 59: here you are defining the path without a trailing slash, and above it is done with it.

VictorHarbo (changes requested):

PR #83, opened 2023-11-14
Disc 361 origin cleanup

VictorHarbo:
Clean up clean up clean up

### DIFF
--- a/conf/ds-present-behaviour.yaml
+++ b/conf/ds-present-behaviour.yaml
@@ -112,7 +112,7 @@ config:
 
   # Collections are to be set in ds-present-kb-collections-yaml if possible, as that file overwrites the collections defined here
   # The collections available for this instance of ds-present
-  #collections:
+  #origins:
     # General note: All collections are expected to support at least the views raw, mods, jsonld and solrjson.
     # If any of those are not supported, the transformer 'fail' should be used with a telling message.
 
--- a/conf/ds-present-behaviour.yaml
+++ b/conf/ds-present-behaviour.yaml
@@ -102,7 +102,7 @@ config:
       # and the second being the collection specific par of the ID
       # Mandatory. Suggested value: '([a-z0-9.]+):([a-zA-Z0-9:._-]+)', e.g. 'images.dsfl:Image1234:2_3-b'
       pattern: '([a-z0-9.]+):([a-zA-Z0-9:._-]+)'
-  collection:
+  origin:
     prefix:
       # Pattern for acceptable collection prefixes. This will practically always be a mirror of the first capturing
       # group for record.id.pattern.
--- a/conf/ds-present-behaviour.yaml
+++ b/conf/ds-present-behaviour.yaml
@@ -98,30 +98,30 @@ config:
   # Settings for handling at the record level
   record:
     id:
-      # Pattern for acceptable record IDs. Must contain exactly 2 capturing group, the first being the collection prefix
-      # and the second being the collection specific par of the ID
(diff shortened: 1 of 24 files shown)


conf/ds-present-behaviour.yaml
```diff
@@ -112,7 +112,7 @@ config:
 
   # Collections are to be set in ds-present-kb-collections-yaml if possible, as that file overwrites the collections defined here
   # The collections available for this instance of ds-present
-  #collections:
+  #origins:
```
tokee, line 115: The comments should also be changed to use `origin`.

conf/ds-present-kb-collections.yaml
```diff
@@ -42,7 +42,7 @@ config:
                 - old_imageserver: ${path:config.oldimageserver.url}
 
   # The collections available for this instance of ds-present
-  collections:
+  origins:
```
tokee, line 45: The file itself (`ds-present-kb-collections.yaml`) should also be renamed.

src/main/java/dk/kb/present/StorageHandler.java
```diff
@@ -14,22 +14,19 @@
  */
 package dk.kb.present;
 
+import dk.kb.present.storage.DSStorage;
 import dk.kb.present.storage.Storage;
 import dk.kb.present.storage.StorageController;
 import dk.kb.util.yaml.YAML;
-import javassist.NotFoundException;
```
tokee, line 20: What was that and how did it ever work!?

src/main/java/dk/kb/present/StorageHandler.java
```diff
@@ -14,22 +14,19 @@
...
-import javax.management.ServiceNotFoundException;
-import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
 /**
  * Creates {@link DSStorage}s from a given configuration. Storages are used for retrieving raw records by
- * {@link DSCollection} and can be shared between multiple collections.
+ * {@link DSOrigin} and can be shared between multiple originss.
```
tokee, line 29: `originss` → `origins`

src/main/java/dk/kb/present/api/v1/impl/IiifPresentationApiServiceImpl.java
```diff
@@ -22,6 +22,7 @@
 import dk.kb.present.model.v1.ManifestServicesDto;
 import dk.kb.present.model.v1.ManifestStartDto;
 import dk.kb.present.model.v1.ManifestThumbnailDto;
+import dk.kb.present.model.v1.OriginDto;
```
tokee, line 25: Not used. But in general we have a bad habit of keeping unused imports.,,

tokee (changes requested):
We thank the developers of IntelliJ IDEA for their skills at making refactoring work. Minor comments and you only missed two spots (that I could find):

* The `View`-constructor takes `originOfCollection`, which now is simply `origin`
* The Javadoc for the `stripPrefix` in `FileStorage`-constructor

src/main/java/dk/kb/present/StorageHandler.java
VictorHarbo: wauw! No clue!

conf/ds-present-behaviour.yaml
tokee: I took the liberty of fixing the remaining two old-name-in-comments.

PR #84, opened 2023-11-16
Disc 427 empty fields

VictorHarbo:
Corrections of XSLTs and solr schema

### DIFF
--- a/src/main/resources/xslt/schemaorg2solr.xsl
+++ b/src/main/resources/xslt/schemaorg2solr.xsl
@@ -450,6 +450,45 @@
         <xsl:value-of select="$internalMap('kb:program_structure_overlaps')"/>
       </f:string>
     </xsl:if>
+
+    <!-- Extracting overlaps sorted by type of overlap as that is the only value to distinguish overlaps by. -->
+    <xsl:if test="f:exists($internalMap('kb:program_structure_overlap'))">
+      <xsl:variable name="overlaps" as="item()*">
+        <xsl:copy-of select="array:flatten($internalMap('kb:program_structure_overlap'))"/>
+      </xsl:variable>
+
+      <!-- Defines fields for overlaps of type 1 and 2. Currently, I don't know if more than two types exists, '
+           then they would have to be added. -->
+      <xsl:for-each select="$overlaps">
+        <xsl:choose>
+          <xsl:when test="map:get(. , 'overlap_type') = '1'">
+            <f:string key="internal_program_structure_overlap_type_one_length_ms">
+              <xsl:value-of select="map:get(. , 'overlap_length')"/>
+            </f:string>
+            <f:string key="internal_program_structure_overlap_type_one_file1UUID">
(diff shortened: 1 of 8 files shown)


src/main/resources/xslt/schemaorg2solr.xsl
```diff
@@ -450,6 +450,45 @@
...
+      <xsl:variable name="overlaps" as="item()*">
+        <xsl:copy-of select="array:flatten($internalMap('kb:program_structure_overlap'))"/>
+      </xsl:variable>
+
+      <!-- Defines fields for overlaps of type 1 and 2. Currently, I don't know if more than two types exists, '
+           then they would have to be added. -->
+      <xsl:for-each select="$overlaps">
+        <xsl:choose>
+          <xsl:when test="map:get(. , 'overlap_type') = '1'">
+            <f:string key="internal_program_structure_overlap_type_one_length_ms">
```
tokee, line 465: `internal_program_structure_overlap_type_one_length_ms` :grimacing: 
But okay, I get that we need to spell it put. Keep it as-is.

src/main/solr/dssolr/conf/schema.xml
```diff
@@ -661,10 +661,33 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?description ?>
     <?example     ?>
   </field>
-  <!--TODO: Figure out what this field can contain-->
-  <field name="internal_program_structure_overlaps" type="string">
-    <?description ?>
-    <?example     ?>
+  <field name="internal_program_structure_overlaps" type="boolean">
+    <?description A boolean value that indicates if there are overlaps between this resource and other resources.?>
```
tokee, line 665: What is an overlap? I'm guessing it is time based and probably coupled to wall time? Don't be afraid to use more words!

src/main/solr/dssolr/conf/schema.xml
```diff
@@ -661,10 +661,33 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
...
+    <?example     true?>
+  </field>
+  <field name="internal_program_structure_overlap_type_one_length_ms" type="plong">
+    <?description Indicates the length of an overlap of type one in ms.?>
```
tokee, line 669: What is type one? As we don't know, please make it easier for future developers to investigate by stating what we /do/ know, i.e. that it is from the Preservica system.

tokee (commented):
Looks like a job well done. Only minor comments about documentation.

src/main/resources/xslt/schemaorg2solr.xsl
VictorHarbo: Not pretty. Lets remember to define which fields that are to be returned in a "normal" solr query.

src/main/solr/dssolr/conf/schema.xml
tokee: Very nice now!

PR #86, opened 2023-11-20
Disc 410 recordsraw endpoint

VictorHarbo:


src/main/java/dk/kb/present/DSOrigin.java
```diff
@@ -198,6 +198,49 @@ public Stream<DsRecordDto> getDSRecords(
...
+                    .flatMap(accessFilter)
+                    .peek(record -> {
+                        try {
+                            record.data(view.apply(record));
+                        } catch (Exception e) {
+                            throw new RuntimeTransformerException(
+                                    "Exception transforming record '" + record.getId() + "' to format '" + format + "'");
+
+                        }
+                    });
```
tokee, line 235: We have the same (ugly & made by me) piece of code below. Could you extract a `safeView`-method that takes care of the transformation and the exception handling?

src/main/java/dk/kb/present/api/v1/impl/DsPresentApiServiceImpl.java
```diff
@@ -149,4 +149,28 @@ public StreamingOutput getRecords(String origin, Long mTime, Long maxRecords, St
         }
     }
 
+    @Override
+    public StreamingOutput getRawRecords(String origin, Long mTime, Long maxRecords, Boolean asJsonLines) {
```
tokee, line 153: A method named `getRawRecords` that calles another method named `getRecordsRaw` :-)

src/main/openapi/ds-present-openapi_v1.yaml
```diff
@@ -172,6 +170,60 @@ paths:
...
+
+  /recordsraw:
+    get:
+      tags:
+        - '${project.name}'
+      summary: 'Extract raw records from a collection after a given mTime.'
+      description: |
+        The extracted records are returned in sorted order by mTime increasing. By changing the parameter `maxRecords`, 
+        the size of the result can be defined. Records marked with a delete flag will also be returned. Records are returned 
+        in a raw storage record.
```
tokee, line 182: Or more precisely: The raw records are returned wrapped in JSON. The caller does not necessarily know what raw storage format is.

src/main/java/dk/kb/present/PresentFacade.java
```diff
@@ -134,23 +134,10 @@ public static StreamingOutput getRecords(
...
-            Set<String> allowedIDs = new HashSet<>(accessChecker.apply(allIDs));
-            return records.size() == allowedIDs.size() ?
-                    records.stream() : // No need for checking as all IDs passed
-                    records.stream().filter(record -> allowedIDs.contains(record.getId()));
-        };
+        Function<List<DsRecordDto>, Stream<DsRecordDto>> accessFilter = validateAccessForRecords(originID, accessChecker, origin);
 
-        // enum:  ['JSON-LD', 'JSON-LD-Lines', 'MODS', 'SolrJSON', "StorageRecord"]
+        // Maybe re-write this switch to use an actual ENUM?
+        // enum:  ['JSON-LD', 'JSON-LD-Lines', 'MODS', 'SolrJSON']
```
tokee, line 140: Oh yes. Except that I'm fairly sure that `-` is nor allowed in enum IDs, so we need to make an enum where the String representation differs from the IDs. Not a problem, just a note.

tokee (changes requested):
Overall looks good. Minor suggestions inline. And the unit tests needed an aegis refresh. Gotta fix that missing overwrite-thing in aegis.

src/main/java/dk/kb/present/PresentFacade.java
tokee: We'll do that at a later time.

PR #87, opened 2023-11-21
Disc 460 pvica to schema

VictorHarbo:
Refactoring of preservica2schemaorg transformation. 

The transformation now takes in to account, what type of metadata record is in hand before transforming and creating fields. 
This means, that fields like aspect_ratio and color now aren't created for a radio record.

### DIFF
--- a/src/main/resources/xslt/preservica2schemaorg.xsl
+++ b/src/main/resources/xslt/preservica2schemaorg.xsl
@@ -23,6 +23,8 @@
   <xsl:param name="recordID"/>
   <xsl:param name="manifestation"/>
 
+  <!-- MAIN TEMPLATE. This template delegates, which fields are to be created for each schema.org object.
+       Currently, the template handles transformations from Preservica records to SCHEMA.ORG VideoObjects and AudioObjects. -->
   <xsl:template match="/">
     <!-- Saves all extensions in a variable used to check if one or more conditions are met in any of them.
          This is done to create one nested object in the JSON with values from multiple PBC extensions. -->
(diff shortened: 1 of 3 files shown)


src/main/resources/xslt/preservica2schemaorg.xsl
```diff
@@ -40,69 +42,168 @@
...
+            <xsl:with-param name="pbcExtensions" select="$pbcExtensions"/>
+          </xsl:call-template>
+        </xsl:when>
+        <xsl:otherwise>
+          <xsl:call-template name="generic-transformation">
+            <xsl:with-param name="type" select="$type"/>
+            <xsl:with-param name="pbcExtensions" select="$pbcExtensions"/>
+          </xsl:call-template>
+        </xsl:otherwise>
+      </xsl:choose>
```
tokee, line 66: It looks like both `VideoObject` and `AudioObject` calls the same templates as `generic-transformation` + some extra? Couldn't `generic-transformation` be called for everything and then the video/audio-specific parts could be called?

tokee (changes requested):
As usual I cannot vouch for the concrete fields as I do not have the knowledge. Only comment is about "code" re-use.

src/main/resources/xslt/preservica2schemaorg.xsl
VictorHarbo: AudioObject can actually call it directly! As I see it, the video Object cannot make use of it as some of the values are represented in a map that gets closed in to generic-template. I could leave this open to make it possible to add to it. However, that would require, that all templates remembers to close that map after calling the generic-template

PR #89, opened 2023-11-22
Disc 455 and 469 new field and title

VictorHarbo:
This PR contains two DISC issues. 455 and 469. 

Please have an opinion on solr field names and how the value of conditionsOfAccess should be included and what it should be as of now

### DIFF
--- a/src/main/java/dk/kb/present/View.java
+++ b/src/main/java/dk/kb/present/View.java
@@ -154,6 +154,7 @@ public class View extends ArrayList<DSTransformer> implements Function<DsRecordD
         final Map<String, String> metadata = new HashMap<>();
         metadata.put("recordID", record.getId());
         metadata.put("origin", origin);
+        metadata.put("conditionsOfAccess", "placeholderCondition");
         return metadata;
     }
 
--- a/src/main/java/dk/kb/present/View.java
+++ b/src/main/java/dk/kb/present/View.java
@@ -154,7 +154,8 @@ public class View extends ArrayList<DSTransformer> implements Function<DsRecordD
         final Map<String, String> metadata = new HashMap<>();
         metadata.put("recordID", record.getId());
         metadata.put("origin", origin);
-        metadata.put("conditionsOfAccess", "placeholderCondition");
+        //TODO: Update placeholder when actual value is in place
+        metadata.put("conditionsOfAccess", "TODO: placeholderCondition");
         return metadata;
     }
 
--- a/src/main/resources/xslt/preservica2schemaorg.xsl
+++ b/src/main/resources/xslt/preservica2schemaorg.xsl
@@ -220,38 +220,55 @@
 
     <!-- Publisher extraction. Some metadata has two pbcorePublisher/publisher/publisherRole.
          We use the one with the value "kanalnavn" as this should be present in all metadata files.-->
-    <xsl:for-each select="pbcorePublisher">
-      <xsl:if test="./publisherRole = 'kanalnavn'">
-        <f:map key="publication">
-          <f:string key="@type">BroadcastEvent</f:string>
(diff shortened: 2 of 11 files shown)


src/main/java/dk/kb/present/View.java
```diff
@@ -154,6 +154,7 @@ private Map<String,String> createBasicMetadataMap(DsRecordDto record) {
         final Map<String, String> metadata = new HashMap<>();
         metadata.put("recordID", record.getId());
         metadata.put("origin", origin);
+        metadata.put("conditionsOfAccess", "placeholderCondition");
```
tokee, line 157: Nice placeholder. Maybe add a TODO to show that it is intentional to have it here until circumstances change?

src/main/resources/xslt/preservica2schemaorg.xsl
```diff
@@ -220,38 +224,76 @@
...
+                  <f:string key="legalName">TV 2 Danmark</f:string>
+                </f:map>
+              </xsl:when>
+              <xsl:when test="f:starts-with($publisherGeneral, 'tv3')">
+                <f:map key="broadcaster">
+                  <f:string key="@type">Organization</f:string>
+                  <f:string key="legalName">TV3</f:string>
+                </f:map>
+              </xsl:when>
+              <xsl:otherwise></xsl:otherwise>
```
tokee, line 287: No `Unknown` here?

src/main/solr/dssolr/conf/schema.xml
```diff
@@ -435,6 +440,23 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?example     Kagerester?>
   </field>
 
+  <field name="broadcaster" type="string">
```
tokee, line 443: As `string` does not help with search (no normalisation and such), we should have a `copyField` from `broadcaster` to `creator_affiliation_generic`.

src/test/java/dk/kb/present/TestFiles.java
```diff
@@ -28,6 +28,7 @@ public class TestFiles {
     public static final String PVICA_RECORD_e683b0b8 = "internal_test_files/tvMetadata/e683b0b8-425b-45aa-be86-78ac2b4ef0ca.xml";
     // White program radio metadata from 1966
     public static final String PVICA_RECORD_c295ae6c = "internal_test_files/tvMetadata/c295ae6c-fd34-4694-a9cc-4204b4a9f3a0.xml";
+    public static final String PVICA_RECORD_accf8d1c = "internal_test_files/tvMetadata/accf8d1c-3c4f-41a9-a1a2-48ce1cf0c5ca.xml";
```
tokee, line 31: It's not a decent pull request if there hasn't been at least one addition or update to `aegis` :grin:

tokee (changes requested):
Minor issues only, except that `PresentFacadeTest.accessFilterMultiRecords` does not pass.

src/main/resources/xslt/preservica2schemaorg.xsl
VictorHarbo: In schema.org we don't really use unknowns. It's more common to omit the field then.

src/main/solr/dssolr/conf/schema.xml
VictorHarbo: I see your point, however the values in `broadcaster` and `creator_affiliation_generic` are different .`broadcaster` is an official organisation name like 'DR' or 'TV 2 Danmark'. While  `creator_affiliation_generic` is a shorthand for a channel like 'drp1'

tokee: Okay, then this will be part of the overarching issue of how we handle fields in Solr. Postponed for later.

PR #92, opened 2023-11-23
Add option for skipping failing records during export

tokee:
See subject


VictorHarbo (approved):
LGTM! I've just added a solr dependency.

PR #94, opened 2023-11-29
Add continuation support

tokee:
Adding continuation stream support to `ds-present` was not trivial as `ds-present` has its own storage abstraction (primarily to support testing) which had do be adjusted to deliver continuation data. This was not possible for `MultiStorage`, which we do not currently use.
As with`ds-storage`, integration tests has been added (see `DsPresentClientTest`). These rely on a matching `ds-present` being deployed on the devel server (this has been done) and on on a `kb init`.


thomasegense (approved):
Merged

PR #96, opened 2023-11-29
Make PresentFacadeTest.accessFilterMultiRecords deterministic

tokee:
The old unit test for access filter was flaky as the result depended on the number of test files. The new code only allows 3 specific records.


VictorHarbo (approved):
LGTM!

PR #97, opened 2023-11-29
Disc 501 remove map find

VictorHarbo:
Instead of using the map:find function to find key entries in nested JSON, I've introduced some safer functions to ilook into nested json. Furthermore I have moved these functions to a utils XSLT to clean up the actual transformer.

### DIFF
--- a/src/main/resources/xslt/schemaorg2solr.xsl
+++ b/src/main/resources/xslt/schemaorg2solr.xsl
@@ -12,6 +12,7 @@
   
   <xsl:output method="text" />
   <xsl:param name="schemaorgjson"/>
+  <xsl:include href="src/main/resources/xslt/utils.xsl"/>
 
   <!--Saves the input JSON as an XDM object. -->
   <xsl:variable name="schemaorg-xml" as="item()*">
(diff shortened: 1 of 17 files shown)


src/main/resources/xslt/schemaorg2solr.xsl
```diff
@@ -97,31 +98,31 @@
...
+                  <xsl:value-of select="my:getTripleNestedMapValue($schemaorg-xml, 'publication', 'publishedOn', 'alternateName')"/>
+                </f:string>
+              </xsl:if>
+
+              <xsl:if test="not(f:empty(my:getFourLevelNestedMapValue($schemaorg-xml, 'publication',
+                                                        'publishedOn', 'broadcaster', 'legalName')))">
+                <f:string key="broadcaster">
+                  <xsl:value-of select="my:getFourLevelNestedMapValue($schemaorg-xml, 'publication',
+                                                          'publishedOn', 'broadcaster', 'legalName')"/>
                 </f:string>
```
tokee, line 121: Would it be possible to extend the helper methods even further so that they also do the "if exists then set the key"-step as that if a very recurring pattern?

tokee (changes requested):
Much clearer with the extraction to util. Maybe room for extra improvement? (see comments)

src/main/resources/xslt/schemaorg2solr.xsl
VictorHarbo: It would only be possible to return the value as they already do. This is because the key attribute on strings, arrays etc. doesn't accept variables. We have to define the key manually. I could create quite specific if statements to handle this, but I don't think that would make anything prettier. It would just move the mess to util.xsl

PR #98, opened 2023-11-29
Disc 531 intermediate solution overlaps

VictorHarbo:
Implements the intermediate solution for representing overlaps in solr documents.

### DIFF
--- a/src/main/solr/dssolr/conf/schema.xml
+++ b/src/main/solr/dssolr/conf/schema.xml
@@ -732,6 +732,14 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?description The UUID of the second file that is part of overlap one.?>
     <?example     f73b69da-2bc0-4e06-b19b-95f24756804e?>
   </field>
+  <field name="internal_overlapping_files" type="string" multiValued="true">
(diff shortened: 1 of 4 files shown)


src/main/solr/dssolr/conf/schema.xml
```diff
@@ -731,6 +733,14 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
   <field name="internal_program_structure_overlap_type_two_file2UUID" type="string">
     <?description The UUID of the second file that is part of overlap one.?>
     <?example     f73b69da-2bc0-4e06-b19b-95f24756804e?>
+  </field>-->
+  <field name="internal_overlapping_files" type="string" multiValued="true">
+    <?description If there are overlaps between files related to a metadata resource. the UUIDs of the relevant files
+                  will be listed in this field. These overlaps can be of all types and there is currently no possibility
+                  for knowing how long the overlap is or what kind of overlap there are between the files. Overlapping
+                  files are listed in pairs, which means that the first and second entry in the field are overlapping
```
tokee, line 741: I like the idea of listing as pairs, but it seems to me that they are just added as independent values in the multi valued field?

tokee (changes requested):
The values does not seem to be paired?

PR #101, opened 2023-11-30
Improve stopOnError

tokee:
The old handling of single record transformation errors was poorly implemented, both regarding code logic and log message. This pull request cleans up the code and makes it explicit in the log that the error is not fatal.

### DIFF
--- a/src/test/java/dk/kb/present/PresentFacadeTest.java
+++ b/src/test/java/dk/kb/present/PresentFacadeTest.java
@@ -16,6 +16,7 @@ package dk.kb.present;
 
 import dk.kb.present.config.ServiceConfig;
 import dk.kb.util.Resolver;
+import org.apache.commons.io.IOUtils;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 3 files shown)


src/test/java/dk/kb/present/PresentFacadeTest.java
```diff
@@ -160,17 +180,14 @@ void getRecordsJSONLD() throws IOException {
...
     @Test
     void getRecordsJSONLDLines() throws IOException {
-        StreamingOutput out = PresentFacade.getRecords(null, "dsfl", 0L, -1L, "json-ld-lines");
+        StreamingOutput out = PresentFacade.getRecords(null, "dsfl", 0L, -1L, "json-ld-lines", ids -> ids);
         String result = toString(out);
         assertTrue(result.contains("\"@value\":\"Letters to\\/from David Simonsen\"}"), "Result should contain the name David Simonsen in the expected wrapping");
         assertFalse(result.contains(",\n"), "Result should not contain a comma followed by newline as it should be a multi-entry JSON-Lines");
         assertFalse(result.endsWith("]\n"), "Result should not end with ']' as it should be in JSON-Lin es");
     }
-*/
```
VictorHarbo, line 173: This test is failing. I have no clue on what the comment _Can only be fixed, when the updated XSLT to JSON-LD has been reviewed and merged to master_ refers to. I believe it is REALLY old and isn't related to the XSLTs i've written.

VictorHarbo (changes requested):
One failing unittest related to old test corpus i believe

src/test/java/dk/kb/present/PresentFacadeTest.java
tokee: Removed

VictorHarbo (approved):
LGTM!

PR #102, opened 2023-12-02
remove string manipulation from view

VictorHarbo:


tokee (approved):
Mmmm... regexp. Much cleaner!

PR #103, opened 2023-12-04
suggest request handler

thomasegense:
requesthandler + search compontent for suggest added. 
It only uses the title field

### DIFF
--- a/src/main/solr/dssolr/conf/solrconfig.xml
+++ b/src/main/solr/dssolr/conf/solrconfig.xml
@@ -836,12 +836,12 @@
       <str name="suggest.count">10</str>
       </lst>
       <arr name="components">
-        <str>dr_title_suggest</str>
+        <str>suggest</str>
       </arr>
     </requestHandler>
 
-  <searchComponent name="dr_title_suggest" class="solr.SuggestComponent">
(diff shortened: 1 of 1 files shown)


src/main/solr/dssolr/conf/solrconfig.xml
```diff
@@ -811,6 +811,50 @@
...
+
+
+<!-- 
+   suggest for title field. If more fields are needed for suggest compontent the best way is to use copy-field and use that field.
+  Important: It is case sensitive now
+  Important: Suggest index will be build on start-up now. Change for large index
+  
+  Example calls:
+  http://devel11:10007/solr/ds/suggest?suggest.dictionary=dr_title_suggest&suggest.q=tv
+  http://devel11:10007/solr/ds/suggest?suggest.dictionary=dr_title_suggest&suggest.q=Kendtes
```
tokee, line 834: Sorry, no internal names allowed. A big shame as I love having demo links in the comments.

tokee (changes requested):
Unfortunately we cannot have internal names in out public repos. Other than that I'm impressed by the comment section - everything is spot on.

tokee (approved):

PR #104, opened 2023-12-05
Disc 300 config cleanup

VictorHarbo:


tokee (approved):
Inspected, deployed & tested.

PR #108, opened 2023-12-11
Disc 540 remove empty map

VictorHarbo:
This PR makes sure, that the map KB:internal isn't created when there are no values to put into it by checking for existence of such values. 

Furthermore the PR updates the XSLTs to make better use of the newly introduced functions for checking nested maps. In this regard a new function has been introduced which makes it possible to check for values of type map in an array by


tokee (approved):
This looks horrible and right. The beauty of XSLT. No corrections.

PR #109, opened 2023-12-12
Prepare for OAuth2 flow

tokee:
The Ds-services are expected to be OAuth2 enabled in the future. This is needed for some endpoints, where special privileges should disable licence checks for exports; needed for the initial update of the Solr index as the Solr index is used by the license module - chicken & egg.

This pull request adds support for the HTTP header
```
Simulated-OAuth2-Group: internal_service
```
when calling web service endpoints. If the header is set, the license check is bypassed. This is **absolutely not** an attempt at providing access checks, but serves to test the flow of headers between our services as well as having different groups of callers.

For debugging, the endpoint `/monitor/whoami` has been added. It can be tested with 
```
curl -X GET "http://localhost:9073/ds-present/v1/probe/whoami" -H  "Simulated-OAuth2-Group: foo"
```

### DIFF
--- a/src/main/java/dk/kb/present/util/DsPresentClient.java
+++ b/src/main/java/dk/kb/present/util/DsPresentClient.java
@@ -261,7 +261,9 @@ public class DsPresentClient extends DsPresentApi {
         if (headers == null || headers.length == 0) {
             return Collections.emptyMap();
         }
+
         return Arrays.stream(headers)
+                .filter(Objects::nonNull)
                 .flatMap(hs -> hs.entrySet().stream())
                 .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
     }
--- a/src/main/java/dk/kb/present/util/DsPresentClient.java
+++ b/src/main/java/dk/kb/present/util/DsPresentClient.java
@@ -72,13 +72,12 @@ public class DsPresentClient extends DsPresentApi {
      * <p>
      * This constructor expects the structure
      * <pre>
-     * config:
-     *   present:
-     *     url: 'http://localhost:9073/ds-present/v1'
-     *     headers:
-     *       # Placeholder for authentication until OAuth2 integration is in place
-     *       # Set this to 'internal_service` on local and devel to bypass license check in ds-present
-     *       - Simulated-OAuth2-Group: anonymous
+     * present:
+     *   url: 'http://localhost:9073/ds-present/v1'
+     *   headers:
+     *     # Placeholder for authentication until OAuth2 integration is in place
+     *     # Set this to 'internal_service` on local and devel to bypass license check in ds-present
+     *     - Simulated-OAuth2-Group: anonymous
      * </pre>
      * @param config following the outlined structure.
      * @param headers zero or more maps of headers, which will be set for all calls by this client.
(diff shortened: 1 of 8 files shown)


src/main/java/dk/kb/present/util/DsPresentClient.java
```diff
@@ -58,40 +63,75 @@ public class DsPresentClient extends DsPresentApi {
...
+
+    /**
+     * Create a client for the remote ds-present service.
      * <p>
-     * When working with YAML configs, it is suggested to define the ds-present URI as the structure
+     * This constructor expects the structure
      * <pre>
-     * storage:
-     *   url: 'http://localhost:9073/ds-storage/v1'
+     * config:
```
VictorHarbo, line 75: Config shouldn't contain the config: key

src/main/java/dk/kb/present/util/DsPresentClient.java
```diff
@@ -58,40 +63,75 @@ public class DsPresentClient extends DsPresentApi {
...
-                setScheme(serviceURI.getScheme()).
-                setHost(serviceURI.getHost()).
-                setPort(serviceURI.getPort()).
-                setBasePath(serviceURI.getRawPath());
+    /**
+     * Create a client for the remote ds-present service.
+     * <p>
+     * When working with YAML configs, it is suggested to use the structure
+     * <pre>
+     * config:
```
VictorHarbo, line 110: Same comment as above

src/main/java/dk/kb/present/webservice/AccessUtil.java
```diff
@@ -35,6 +36,10 @@
 public class AccessUtil {
     private static final Logger log = LoggerFactory.getLogger(AccessUtil.class);
 
+    public static final String HEADER_SIMULATED_GROUP = "Simulated-OAuth2-Group";
+    public static final String GROUP_INTERNAL_SERVICE = "internal_service";
+    public static final String GROUP_ADMIN = "admin";
```
VictorHarbo, line 41: Question: How have these values been chosen?

src/main/openapi/ds-present-openapi_v1.yaml
```diff
@@ -377,6 +383,20 @@ paths:
...
+        - Service
+      summary: 'Pseudo-OAuth2 check: The groups that the caller belongs'
+      operationId: whoami
+      responses:
+        '200':
+          description: 'OK'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Whoami'
```
VictorHarbo, line 399: I like this endpoint, works like a charm

VictorHarbo (changes requested):
Code looks good. Minor comments. Also, a test fails after updating aegis and running kb init:  DsPresentClientTest.testRemoteRecordsStream:88 The requested number of records should be received ==> expected: <3> but was: <0>

src/main/java/dk/kb/present/webservice/AccessUtil.java
tokee: By careful deliberation and pondering for about 5 seconds.

They will be replaced later on. I'll add a comment about it to the code.

VictorHarbo (approved):
LGTM!

PR #112, opened 2023-12-18
Disc 505 xslt errorhandling

VictorHarbo:
Add try-catch to XSLTs. I haven't figured how to create a unittest for it. To see the functionality in action you could add the following inside a try-tag and then run the following tests `dk.kb.present.transform.XSLTPreservicaSchemaOrgTransformerTest#testSetup` and `dk.kb.present.transform.XSLTPreservicaToSolrTransformerTest#prettyPrintTransformation`:

 `<f:string key="tester">
    <xsl:value-of select="20 div 0"/>
  </f:string>`


tokee (commented):
I would really like a unit test for this! I understand the problem: Anywhere a fault could be triggered by creating special input data, a more proper action would be to ensure the fault cannot be triggered.

So we're looking for something where "this will **always** be so for a real input, so we don't need to check...
What about the `manifestationRef` in `preservice2schemaorg.xml` line 579? If that was shorter than 5, would it trigger an error?

VictorHarbo:
I fiddled with your suggestion and I found what I believe is a prettier way of achieving a working unittest. Instead of giving the transformer an actual XML as manifestation the test gives it at string and fails as wanted.

PR #113, opened 2023-12-19
Disc 575 modified time in solr

VictorHarbo:


src/main/java/dk/kb/present/View.java
```diff
@@ -159,14 +163,35 @@ public String apply(DsRecordDto record) {
...
+     */
+    private static String getSolrDate(String dateTime) {
+        // Define the formatter for the input date string. This is almost already the expected format of ISO 8601.
+        // However, a small conversion needs to be make to change the space in the input pattern to a T.
+        DateTimeFormatter dsStoragePattern = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ssZ", Locale.ROOT);
+        OffsetDateTime offsetDateTime = OffsetDateTime.parse(Objects.requireNonNull(dateTime), dsStoragePattern);
+        Instant instant = offsetDateTime.toInstant();
+        // Convert the datetime string to the ISO INSTANT format, which is in use in both SCHEMA.ORG and SOLR
+        // https://schema.org/DateTime and https://solr.apache.org/guide/6_6/working-with-dates.html#WorkingwithDates-DateFormatting
+        DateTimeFormatter isoInstantFormatter = DateTimeFormatter.ISO_INSTANT;
```
tokee, line 191: Solr supports millisecond granularity: https://solr.apache.org/guide/6_6/working-with-dates.html

I suggest we use the "non-human" `mTime` (which is microsecond granularity) as input instead of `mTimeHuman`.

src/main/solr/dssolr/conf/schema.xml
```diff
@@ -582,6 +582,11 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?example     www.example.com/streaming/bart-access-copies-tv/00/00/00/000000e1-0000-462a-a2b4-7488244fcca7/playlist.m3u8"?>
   </field>
 
+  <field name="date_modified" type="pdate" stored="true">
+    <?description Milliseconds since 1st of January 1970 (epoch) with three added digits representing the last time a record has been modified?>
```
tokee, line 586: That mighe be true beneath the hood, but conceptually it is now simply a point in time (as for `startTime` and `endTime`. And the three digits are gone :-)

Also, while the example is technically correct as a rendering of the `pdate`, so is "November 29., 2023, 6:45PM". It is just a representation.

Solr has no explicit concept of timezones, i.e. all time is Zulu time. That is why the default rendering, the one returned from the common Solr request, is `2023-11-29 14:45:49Z`. The example should follow that pattern.

src/main/solr/dssolr/conf/schema.xml
```diff
@@ -582,6 +582,11 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?example     www.example.com/streaming/bart-access-copies-tv/00/00/00/000000e1-0000-462a-a2b4-7488244fcca7/playlist.m3u8"?>
   </field>
 
+  <field name="date_modified" type="pdate" stored="true">
```
tokee, line 585: This will be changed to something like `internal_mTime` or something like that, right?

src/main/java/dk/kb/present/View.java
tokee: Addendum: We should consider if we want to keep the whole Epoch microsecond-thing in `ds-storage`. It is an uncommon measurement and maybe the price (a maximum of 1000 records ingested per second) for shifting to Epoch milliseconds is acceptible?

tokee (changes requested):
This looks like your usual solid work with nice unit tests. Sorry to be greedy, but I suggest a change to millisecond granularity, which means that you will have to rewrite most of it :-(

src/main/solr/dssolr/conf/schema.xml
VictorHarbo: Yes, will change to probably something like internal_storage_mTime to keep the  internal_ prefix and indicate, that the value originates from our internal ds-storage and not say preservica. This value will also be included in the internal map of the schema.org JSON, while the dateModified there should be reserved for a value related to metadata modification extracted from origin

src/main/java/dk/kb/present/View.java
VictorHarbo: The method getSolrDate isn't needed in the new implementation, however I haven't removed it because it will be needed again, when we are to include a last modified in origin date.

tokee (approved):
Looks good now, merging!

PR #115, opened 2023-12-21
Disc 546 fields for temporal search

VictorHarbo:
First try at implementing solr fields for temporal search. 

The values are not represented in schema.org yet. Do we think they should? 

Please have an extra look on the creation and use of the fields `temporal_start_time_da_date` and `temporal_end_time_da_date` as these fields are quite rule bending (by not being valid dates and in a wrong timezone)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Added [1.4.0]
+- Add solr fields for temporal searching. 
+
 ### Added [1.3.0]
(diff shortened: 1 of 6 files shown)


tokee (changes requested):
The hack where the local Danish time is stored in a `pdate` is acceptable as it is the only known way to do support date math in Solr, as Solr does not have a concept of time zones other than Zulu time.

Unfortunately the conversion does not take summer/winter time into account. I have adjusted one unit test so that it tests properly (and fails) and added another unit test that demonstrates proper conversion. The data formatting code need to be adjusted so that all unit tests passes - I am not sure that can be done entirely in XSLT :disappointed:

VictorHarbo:
Turns out this can be done in XSLT 3.0 by using the five argument function: [format-datetime()](https://www.saxonica.com/html/documentation12/functions/fn/format-dateTime.html)

tokee (approved):
Nice fix (note: the Overton window for "nice" is severely skewer for all XSLT work).

I took the liberty of removing a now-outdated comment on a failing unit test for summer/winter-time.

PR #117, opened 2023-12-28
Disc 431 count fields

VictorHarbo:
Add solr fields for title and description lengths etc. These fields were already created when we worked with 'Billedesamlingen'. However, We haven't until now added them for Radio and TV yet.

### DIFF
--- a/src/test/java/dk/kb/present/transform/XSLTPreservicaToSolrTransformerTest.java
+++ b/src/test/java/dk/kb/present/transform/XSLTPreservicaToSolrTransformerTest.java
@@ -348,8 +348,23 @@ public class XSLTPreservicaToSolrTransformerTest extends XSLTTransformerTestBase
     }
 
     @Test
+    void testCountFields(){
+        assertPvicaContains(TestFiles.PVICA_RECORD_a8aafb121, "\"categories_count\":\"2\"");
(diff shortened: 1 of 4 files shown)


src/test/java/dk/kb/present/transform/XSLTPreservicaToSolrTransformerTest.java
```diff
@@ -347,9 +347,24 @@ void testNoSpecificOrigin(){
...
+        assertPvicaContains(TestFiles.PVICA_RECORD_a8aafb121, "\"notes_count\":\"2\"");
+    }
+
+    @Test
+    void testLengthFields(){
+        assertPvicaContains(TestFiles.PVICA_RECORD_a8aafb121, "\"title_length\":\"25\"");
+        assertPvicaContains(TestFiles.PVICA_RECORD_a8aafb121, "\"notes_length\":\"389\"");
+        assertPvicaContains(TestFiles.PVICA_RECORD_a8aafb121, "\"abstract_length\":\"15\"");
+        assertPvicaContains(TestFiles.PVICA_RECORD_a8aafb121, "\"description_length\":\"374\"");
+
```
tokee, line 362: Thank you for your unit test discipline!

tokee (approved):
Looks good. There is an unrelated problem with the `DsPresentClientTest` that gets a HTTP 500 from the devel server. Unfortunately logging currently does not work on devel, so it is not obvious what goes wrong.

PR #119, opened 2024-01-10
Disc 628 io records transformation error and DISC-624 empty genre_sub field

VictorHarbo:
I squashed these two issues into the same PR as the code changes where quite small.

- Some IO records have their pid handle twice. These gets compared now and one is extracted. 
- As @thomasegense noted there are a lot of records with the field `genre_sub` with the value being an empty string. This is also fixed.
 
I'll also create an issue on going through the index looking for more fields with empty values as I suspect that I might have introduced these with my JSON parsing XSLT functions.


tokee (approved):
Looks good and passes unit tests. Some sort of run-through-the-map-recursively-and-remove-blanks post-processing step would be nice to avoid this kind of errors in the future.

PR #120, opened 2024-01-10
Disc 587 suggest dictionary

VictorHarbo:
Default suggest dictionary has been added to config. 

I had some problems getting the suggest component working on devel. This was due to it not being build when new collections were created. It was correctly build when solr was restarted and when the configuration was updated. When the component haven't been build it can be build during query by adding `&suggest.build=true` to the query. I've added a note on this in our setup documentation

### DIFF
--- a/src/main/solr/dssolr/conf/schema.xml
+++ b/src/main/solr/dssolr/conf/schema.xml
@@ -87,7 +87,7 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
     <?query_example ?>
   </field>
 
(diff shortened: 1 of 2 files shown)


thomasegense (approved):
Den virker (stadig...)
http://devel11:10007/solr/ds/suggest?suggest.dictionary=dr_title_suggest&suggest.q=tv

Forstår ikke hvordan dette er gået tabt.
Der mangler måske lige en default hvis man ikke angiver suggest.dictionary. Det var også lavet sidst, men det kan jeg bare smide på.

PR #121, opened 2024-01-12
Solr configuration update

tokee:
Updating the configuration for an existing Solr Cloud 9.4 collection can be done three different ways, all officially supported:

* Using ZooKeeper's `linkconfig`
* Using Solr's `collections` API v1 (plain REST GET params @ `/solr/admin/collections`)
* Using Solr's `collections` API v2 (JSON POSTed to `/api/collections/<collection_name>`)

First one fails silently, second one fails with a ClassCastException, third one works! I should look into this and possibly file a Solr bug report, but that'll be on private time.

This fix makes `bin/cloud_ds.sh update` work. Test with
```shell
bin/cloud_install.sh
bin/cloud_start.sh
bin/cloud_ds.sh new
bin/cloud_ds.sh align
```
Then
* Inspect the schema through `Files` in the Solr admin GUI for `_ds_1.0.3_`
* bump the version in `src/main/solr/dssolr/conf/schema.xml`:
  `<field name="_ds_1.0.3_" type="string" indexed="false" stored="false"/>` →
  `<field name="_ds_1.0.4_" type="string" indexed="false" stored="false"/>`)
* call
  `bin/cloud_ds.sh update`
  
Checking the schema again in the Solr admin GUI (might need a reload) should show the new version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added [1.4.0]
 - Add solr fields for temporal searching. 
 
+### Fixed
+- Configuration update of existing collection now works: bin/cloud_ds.sh update
+
 ### Added [1.3.0]
 - Added ReflectUtil class containing methods used for testing. Using these methods instead of internal mockito methods makes the project not depend on mockitos old internal class.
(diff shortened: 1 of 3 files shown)


VictorHarbo (approved):
Nice digging! Schema version updates flawlessly now!

PR #122, opened 2024-01-19
logback fix

thomasegense:
same same


tokee (approved):
Except for a few randomly inserted blanks, this looks good!

PR #125, opened 2024-01-30
Disc 730 spellcheck

thomasegense:
A very good first attempt. 
Example call and try using variantions of "Carina Jensen"
http://devel11:10007/solr/ds-write/select?indent=true&q.op=AND&q=carina%20Jense&rows=1&facet=false&spellcheck=true

**Spellcheck will only be called if no results are found.**

**Frontend only needs to read the json-attribute`collationQuery:"carina jensen"`**  (Can be several, max 5)

Implementation details:

Added comment to important spellcheck parameters. Tried tweak them to a good value for the current corpus.

Only fields **title** and **description** are copied to the spellcheck field. Description is a big field, so there is redundant data in 
the index. But it will not cost performance for searches that have  results.

**ds-discover** needs to accept the spellcheck parameter without warnings. (new task)

collection **ds** has been alligned with **ds-write**. So it is already active for frontenders.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+Spellcheck compontent added to Solr search. If no results are found, the solr reply will come with query suggestions that will give results.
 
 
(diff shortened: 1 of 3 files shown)


tokee (approved):
The solution looks solid for our use and works well for the queries I experimented with. Thank you for the comments in the configs - they are really helpful for further adjustments.

There are 30 commit messages, zero of them have any value. This is a small pull request, so I'll squash them into a single commit. For a larger pull request this would require a proper rebase.

PR #126, opened 2024-02-05
Disc 766 open api from injection plugin

VictorHarbo:


pom.xml
```diff
@@ -699,6 +699,35 @@
...
+                <YamlResolver>
+                  <yamlType>List</yamlType>
+                  <yamlPath>origins</yamlPath>
+                  <sourceKey>origin</sourceKey>
+                  <destinationKey>origins</destinationKey>
+                  <createEnum>true</createEnum>
+                </YamlResolver>
+                <YamlResolver>
+                  <yamlType>Single-value</yamlType>
+                  <yamlPath>origins[7].'ds.tv'.origin</yamlPath>
```
tokee, line 725: Same problem as with [ds-datahandler](https://github.com/kb-dk/ds-datahandler/pull/45): This is a frail construction. Besides, with `ds.tv` already mentioned in the path, why not just state it directly as a property instead of visiting the YAML?

The solution should be the same as the one we choose for `ds-datahandler`.

PR #127, opened 2024-02-06
Disc 259 better search

thomasegense:
Added synonym filter to the existing fieldType 'text_general' fieldtype (stored)
Defined new fieldType: 'text_general_synonym' (stored)
Add a new field 'title_synonym' of fieldType: 'text_general_synonym' (stored)
The  title_synonym is stored is for highlightning. This will mean a search for 'tvavis' can result in 'tv avisen' as highlighted. You agree with this decision?

Another reason for a new title field is because of ranking. Matching the title field is top ranking priority for dr-arkiv search from what meta-data we have available. The title field is multivalued currently and that will be changed later by Victor. 

See the new file for synonyms:
**dr_synonyms.txt** 
It is not a complete list, just a start.

Field analysis of the new field can be tested here:
http://devel11:10007/solr/#/ds/analysis?analysis.fieldvalue=tvavis&analysis.fieldname=title_synonym&verbose_output=1

Search testing: ('official frontend' has OR as default operator, so does not work).
http://devel11:10001/ds-search/
Try search for :
tvavis 
tv avis
tvavisen
tv-avisen
tv avisen
"tv avisen"

Unit test:
Added simple unittests for the synonyms.


src/test/java/dk/kb/present/solr/EmbeddedSolrFieldAnalyseTest.java
```diff
@@ -127,15 +127,27 @@ public void testTitleStrict() throws SolrServerException, IOException {
...
 	
-
-	private long getCreatorNameStrictResultsForQuery(String query) throws Exception{	    
+	@Test
+	public void testSynonymTest() throws SolrServerException, IOException {
+		addSynonymFieldTestDocuments();
+		assertEquals(1, getTitleSynonymQuery("tvavis"));
+		assertEquals(1, getTitleSynonymQuery("\"tvavis\"")); // "tvavis" in quotes
+		assertEquals(1, getTitleSynonymQuery("tv-avisen")); //"tv-avisen" in quotes
+		assertEquals(1, getTitleSynonymQuery("\"tv-avisen\""));
```
tokee, line 144: Eclipse has been using tabs for indent, both in new code in this test class and for a lot of the old code in this class. Please fix this in the file, so that you @thomasegense will retain git credit for the code.

Also, please tell Eclipse to _never_ use tabs for indent for Java code.

src/test/java/dk/kb/present/solr/EmbeddedSolrFieldAnalyseTest.java
```diff
@@ -196,6 +224,31 @@ private static void addTestDocuments() throws Exception {
...
+	
+
+	
+	private static void addSynonymFieldTestDocuments()  {
+
+	try {
+				SolrInputDocument document = new SolrInputDocument();
+				document.addField("id", "synonym1");
+				document.addField("origin", "ds.test");
+				document.addField("title", "Velkommen til TV avisen");  //Synonym file: tv-avisen, tvavis, tvavisen, tv-avis => tv avisen
```
tokee, line 236: Could you add a document and a test that reverses the synonym flow? Where `title` is `Velkommen til TVAvisen` and the search is for `TV Avisen`?

tokee (changes requested):
We need to decide if synonyms should be enabled for fields directly or if a copy of the field should be made for synonyms (`title` + `title_synonym`).

It seems that the field copy method would allow for _slightly_ better ranking & precision, but having a copy of the field is a very high price to pay for that. Too high, I would say. Are there other reasons for having 2 fields?

NB: Remember the changelog.

src/test/java/dk/kb/present/solr/EmbeddedSolrFieldAnalyseTest.java
thomasegense: configured eclipse again... And reformated file.

src/test/java/dk/kb/present/solr/EmbeddedSolrFieldAnalyseTest.java
thomasegense: Fixed, and the uppercase TVAvisen revealed and error in the fieldtype analyser. lowercase-filer must be before synonyms (and stopwords).

tokee: I might have been unclear here: My point is that synonym handling of multi-words needs to be checked both ways.

1. A search for `tv avisen` should match a document where `tvavisen` is indexed
2. A search for `tvavisen` should match a document where `tv avisen` is indexed

The unit test needs at least 2 documents, one with and one without space in `tv avisen`.

src/main/solr/dssolr/conf/schema.xml
```diff
@@ -969,17 +989,21 @@ In solrconfig.xml the classic schema is defined with this:  <schemaFactory class
...
 
+  <!-- Field to 'catch em all' if not matched in specific fields. Last resort for matching. Ranking will be lower if they are only matched here. -->
    <fieldType name="freetext" class="solr.TextField" multiValued="false" indexed="true" required="false" stored="true">
     <analyzer type="index">
       <tokenizer class="solr.StandardTokenizerFactory"/>
       <charFilter class="solr.MappingCharFilterFactory" mapping="mapping-FoldToASCII.txt"/>
       <filter class="solr.StopFilterFactory" ignoreCase="true" words="lang/stopwords_da.txt" format="snowball"/>
+      <filter class="solr.SynonymGraphFilterFactory" synonyms="lang/dr_synonyms.txt"/>
+       <filter class="solr.FlattenGraphFilterFactory"/> <!-- required on index analyzers after graph filters -->
       <filter class="solr.LowerCaseFilterFactory"/>
```
tokee, line 1000: Shouldn't the lowercase filter come before the synonym filter?

tokee (changes requested):
2 inline comments. Also, please remember to bump the version field at the top of `schema.xml` (`_ds_1.0.3_` -> `_ds_1.1.0_` as these are breaking changes).

src/main/solr/dssolr/conf/schema.xml
thomasegense: Good find. Forgot to also change this fieldtype definition.

src/test/java/dk/kb/present/solr/EmbeddedSolrFieldAnalyseTest.java
thomasegense: Two different DOCs now and unittest

tokee: I don't see an unit test for case 1: Search for `tv avisen` (with space in the query).

thomasegense: Added search in both document for tv avisen and "tv avisen" (quotes)

tokee (approved):

PR #128, opened 2024-02-08
Disc 656 paging record count

VictorHarbo:
Add paging-record-count to ds-present

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>dk.kb.util</groupId>
             <artifactId>kb-util</artifactId>
-            <version>1.4.22</version>
+            <version>1.4.24</version>
             <exclusions>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -670,7 +670,7 @@
                     <useTestScope>true</useTestScope>
                     <!-- https://www.eclipse.org/jetty/documentation/jetty-10/programming-guide/index.html#jetty-run-goal -->
                     <!-- https://stackoverflow.com/questions/12493798/jetty-maven-plugin-using-scan -->
-                    <scan>2</scan>
+                    <scan>0</scan>
```
tokee, line 673: Not complaining, but what is the reason for disabling hot reload?

tokee (commented):
Simple & clean implementation with a fitting unit test. If you update the changelog, we're good.

PR #130, opened 2024-02-13
Disc 803 spellcheck default

thomasegense:
Simple solrconfig changes.

Set spellcheck to "true" also disable "spellcheck.build" on every request.

### DIFF
--- a/src/main/solr/dssolr/conf/solrconfig.xml
+++ b/src/main/solr/dssolr/conf/solrconfig.xml
@@ -753,7 +753,10 @@
        <str name="indent">true</str>
        <str name="df">freetext</str>
        <str name="q.op">AND</str>
-       <str name="defType">edismax</str>
+       <str name="defType">edismax</str>      
+       <str name="spellcheck">true</str>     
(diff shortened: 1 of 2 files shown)


src/main/solr/dssolr/conf/solrconfig.xml
```diff
@@ -820,12 +819,13 @@
       <str name="spellcheck.collateExtendedResults">true</str> > <!-- Probably disable later.But usefull to see what is in data -->
       <str name="spellcheck.maxCollationTries">10</str>
       <str name="spellcheck.maxCollations">5</str>
-      <str name="spellcheck.build">true</str>              
+      <str name="spellcheck.build">false</str>  <!-- Much overhead building it every query. Not recommended and will not scale -->
```
VictorHarbo, line 822: Just a question, when and how does the spellcheck get build? Is it build and rebuild during indexing or do we need to manually build it somehow?

VictorHarbo (approved):
Approved. Just a single question out of curiosity

PR #132, opened 2024-02-20
Disc 835 suggest filter

VictorHarbo:
The result from the suggest component should be filtered in the same was as a normal query is by a fq. Suggest component uses a contextField for this. 

This branch **is not** deployed on devel. If deploying to devel to test it, remember to create a new collection before updating configuration. Remember to build the index for the suggester by appending `&suggest.build=true` to the first query to ensure that the index is build. eg: `http://localhost:10007/solr/ds/suggest?suggest.q=dr&suggest.cfq=DR&suggest.count=10&wt=json&suggest.build=true`

To test the functionality try the following query: `http://localhost:10007/solr/ds/suggest?suggest.q=dr&suggest.cfq=DR&suggest.count=10&wt=json` and observe how only result from DR are returned. Confirm that _dræberen med de 100 dolke_ isn't part of the suggestions and then do the same query at the frontend application and se that this TV2 show is infact there.

### DIFF
--- a/src/main/resources/xslt/preservica2schemaorg.xsl
+++ b/src/main/resources/xslt/preservica2schemaorg.xsl
@@ -727,6 +727,15 @@
       <xsl:value-of select="format-number($mTime, '0')"/>
     </f:string>
 
+    <xsl:if test="$manifestation != ''">
+      <xsl:variable name="manifestationRef">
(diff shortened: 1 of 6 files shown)


VictorHarbo:
Didn't see the XSLT changes where on this branch. @thomasegense and I discussed these yesterday.

thomasegense (approved):
Seems perfect now. Tried unittests.

PR #133, opened 2024-03-05
Marked unitest with integration tag if they require external resources.

thomasegense:
Fail hard if an expected resource is not present when running tests.

### DIFF
--- a/src/test/java/dk/kb/present/OriginHandlerTest.java
+++ b/src/test/java/dk/kb/present/OriginHandlerTest.java
@@ -1,8 +1,12 @@
 package dk.kb.present;
 
+import dk.kb.present.config.ServiceConfig;
 import dk.kb.present.model.v1.FormatDto;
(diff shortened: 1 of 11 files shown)


src/test/java/dk/kb/present/OriginHandlerTest.java
```diff
@@ -52,7 +57,7 @@ void demoPattern() {
         String id = "images.dsfl:luftfoto-sample.xml";
         Matcher matcher = recordIDPattern.matcher(id);
         if (matcher.matches()) {
-            System.out.println("Collection=" + matcher.group(1) + ", material-id=" + matcher.group(2));
+            log.debug("Collection=" + matcher.group(1) + ", material-id=" + matcher.group(2));
```
VictorHarbo, line 60: Nice catches out system.outs as well

VictorHarbo (approved):
As the others from DRA-103 this is working as it should

PR #136, opened 2024-03-08
Dra 105 solr documentation

VictorHarbo:
Move transformation code to ds-present as discussed.


thomasegense (approved):
Implemented as we agreed.

PR #137, opened 2024-03-13
DRA-236. Removed spellcheck.maxCollationTries

thomasegense:
Remove spellcheck.maxCollationTries since it crashed solr when reloading core after configuration change.

Reload core with:
`curl -X POST "http://localhost:8983/api/collections/ds/" -H 'Content-Type: application/json' -d '{"modify":{"config": "dssolr" } }'`

Full description in the Jira Issue.


tokee (approved):
I have tested with and without `maxCollationTries`and can confirm you observations. Good job bughunting!

PR #139, opened 2024-04-03
Dra 210 update injection plugin

VictorHarbo:


tokee (changes requested):
Now that this is truly dynamic, could you add this to the endpoints `/records`, `/recordsraw` and `origin/{id}?

VictorHarbo:
Closing this as discussed in the office. We are injecting variables dynamically now.

PR #140, opened 2024-04-04
Dra 327 dynamic update openapi

VictorHarbo:
Haven't implemented enums. See the comment in YAML spec for explanation


src/main/openapi/ds-present-openapi_v1.yaml
```diff
@@ -194,7 +196,7 @@ paths:
           required: true
           schema:
             type: string
-            example: ${example_origin}
+            example: ${config:origins[0].'ds.radio'.origin}
```
tokee, line 199: Could this be an enum instead?

src/main/openapi/ds-present-openapi_v1.yaml
```diff
@@ -250,7 +252,7 @@ paths:
           schema:
             type: string
             minLength: 1
-            example: ${example_origin}
+            example: ${config:origins[0].'ds.radio'.origin}
```
tokee, line 255: Same as for `/records`: Could this be an enum?

src/main/openapi/ds-present-openapi_v1.yaml
tokee: (same comment for the `/records`-endpoint)

tokee (changes requested):
Request for enums instead of single values when using expansion.

PR #141, opened 2024-04-09
Bugfix: Switch from Jersey to Apache URI Builder to handle parameters containing '{'

tokee:
Jersey's `UriBuilder` treats `{` as the signal for an expansion mechanism, making requests containing `{` fail. This pull request switches to Apache's `URIBuilder`.

### DIFF
--- a/src/main/java/dk/kb/present/util/DsPresentClient.java
+++ b/src/main/java/dk/kb/present/util/DsPresentClient.java
@@ -191,7 +191,7 @@ public class DsPresentClient extends DsPresentApi {
         URI uri;
         try {
             uri = new URIBuilder(serviceURI)
-                    .setPath("records")
+                    .setPath("recordsraw")
                     .addParameter("origin", origin)
                     .addParameter("mTime", Long.toString(mTime == null ? 0L : mTime))
                     .addParameter("maxRecords", Long.toString(maxRecords == null ? 10 : maxRecords))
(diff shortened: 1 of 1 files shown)


src/main/java/dk/kb/present/util/DsPresentClient.java
```diff
@@ -175,13 +188,24 @@ public ContinuationInputStream<Long> getRecordsJSON(String origin, Long mTime, L
...
-                .path("recordsraw")
-                .queryParam("origin", origin)
-                .queryParam("mTime", mTime == null ? 0L : mTime)
-                .queryParam("maxRecords", maxRecords == null ? 10 : maxRecords)
-                .queryParam("asJsonLines", asJsonLines != null && asJsonLines)
-                .build();
+        URI uri;
+        try {
+            uri = new URIBuilder(serviceURI)
+                    .setPath("records")
```
VictorHarbo, line 194: Should be recordsraw right?

VictorHarbo (changes requested):
Small correction needed and please update the changelog as well.

src/main/java/dk/kb/present/util/DsPresentClient.java
tokee: Kinda important to call the right endpoint. Thanks!

VictorHarbo (approved):
Looks good now

PR #143, opened 2024-04-19
 Switch config to camelCase

tokee:
This pull request changes most multi-word configuration keys to camelCase

aegis has an equivalent branch https://github.com/kb-dk/aegis/tree/DRA-431_config_style but the aegis branch is expected to be extended as more DS-services gets updated.

### DIFF
--- a/conf/ds-present-behaviour.yaml
+++ b/conf/ds-present-behaviour.yaml
@@ -100,7 +100,7 @@ storages:
 
 # Settings for handling export of multiple records
 records:
-  errors:
+  errorHandling:
     # If true, export is stopped if a transformation fails for any record
     # If false, failed records are logged and skipped
     # Default: true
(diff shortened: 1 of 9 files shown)


conf/ds-present-behaviour.yaml
```diff
@@ -90,19 +91,20 @@ storages:
...
-  # If true, export is stopped if a transformation fails for any record
-  # If false, failed records are logged and skipped
-  # Default: true
-  stoponerrors: true
-
+  errors:
+    # If true, export is stopped if a transformation fails for any record
+    # If false, failed records are logged and skipped
+    # Default: true
+    stop: true
```
VictorHarbo, line 107: I like that this is more YAML'esque, however, i feel like it is hard to decode the meaning of the path `errors.stop` - would a primary key like errorHandling etc. make more sense?

conf/ds-present-kb-origins.yaml
VictorHarbo, line 1: All of this cleaning up makes my ordered mind so happy!

VictorHarbo (changes requested):
Tiny comment which I'd like your opinion on

VictorHarbo (approved):
Perfect!

PR #145, opened 2024-05-02
Dra 481 prepare solr for stage

VictorHarbo:
Adds an assembly step for delivering the solr config set as a zip file. 
The only thing a bit worrying here is the versioning of the solr schema, which has to be done in two places now, however I don't see another way of doing it in the schema, which is quite important for our `ds_cloud`-script. 

Naming has been done following this internal guide on naming a solr collection for production at KB: https://kb-dk.atlassian.net/wiki/spaces/DT/pages/13829092/Solr+Cloud+som+service


thomasegense (approved):
Can confirm the 'ds_solr9_conf_1.6.3.zip' was created as part of build.

PR #148, opened 2024-05-13
Dra 607 create year field

VictorHarbo:
Implement temporal_start_year field


thomasegense (approved):
Looks good! And you remembered the solr-version both places.

PR #149, opened 2024-05-16
Dra 636 add month field solr

VictorHarbo:
Add month field and make year and month field fq-able by request from @jesperlauridsen.


thomasegense (approved):
LGTM

PR #153, opened 2024-06-04
Dra 763 build error

VictorHarbo:
Comparing with ds-storage, which is running correctly in stage, ds-present was missing the ```slf4j-ext``` dependency, which storage inherits from another dependency. We should discuss if this is to be included directly in all services.


thomasegense (approved):
Lets try it...

PR #154, opened 2024-06-07
Dra 674 remove preservica5 from xslts

VictorHarbo:
Remove old preservica 5 specific code also updating all tests to use preservica 7 data and in that case found some cases, which wasnt handled correctly in the XSLT for preservica 7 data.


thomasegense:
fint tager jeg nu

thomasegense (approved):
Thanks, much needed cleanup!

PR #155, opened 2024-06-12
Dra 831 implement faceting fieldtype

VictorHarbo:
Implements a specific field type used for faceting. Basically a non-stored string.
Also converts all currently specified faceting fields to use this field type by supplying ned fields with a `_facet`-suffix.


thomasegense (approved):
OK

PR #157, opened 2024-06-18
Make the transformations handle basic differences between doms and preservica

VictorHarbo:
Makes the transformations handle basic differences between doms and preservica data. This PR targets and fixes the following:
* DOMS records contain namespace prefixes on each XML element, which preservica native records doesn't. When processing PBCore child elements, the prefix is now stripped from the record
* Datetime objects can be delivered in UTC and UTC+1 timezones from the OAI-PMH harvest. These are now converted to UTC only.

The changes have been tested locally on 5k DOMS records from preservica devel that I've succesfully indexed in a local solr instance. We should wait for the updated preservica stage test before running a new ingest and index on our shared devel.

NB: I've added a JIRA issue on creating a field for recognizing doms records and prettifying the content of the transformed records.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Added XSLT support for parsing Preservica and DOMS records through the same transformation.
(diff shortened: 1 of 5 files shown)


thomasegense (approved):
LGTM.

PR #161, opened 2024-07-10
Move facet field creation to XSLT to handle upper casing

VictorHarbo:
Moved creation of facet field to XSLT instead of making use of copy field. This is done like this to keep the facet field type as optimized as possible. docValues cant be used on a solr.TextField whcich would be needed to handle the casing in the schema.


thomasegense (approved):
LGTM!

PR #162, opened 2024-07-12
Dra 678 field review

VictorHarbo:
After contextual review from Karen there had to be made some changes to already existing fields, plus creation of the new field country_of_origin.

Also, the internal JIRA DRA-1095 has been extended to include more missing fields.


thomasegense (approved):
LGTM!

PR #165, opened 2024-07-17
Dra 948 doms transformation cleanup 

VictorHarbo:
First part of DRA 948 tackling DOMS records being indexed as MediaObjects due to no-namespace prefixes for PbcoreDescriptionDocuments. 

To run tests please do a `kb init`


thomasegense (approved):
ok

PR #166, opened 2024-07-18
Dra 1137 correction of transformations

VictorHarbo:
More XSLT changes catching non-pretty transformed DOMS records. Adds a migrated_from field but mainly prettifies the fields:
- Title/name
- alternateName
- videoFrameSize/aspect_ratio
- countryOfOrigin/country_of_origin


thomasegense (approved):
LGTM!

PR #168, opened 2024-07-22
Dra 436 enable oauth2

thomasegense:
All OAuth classes and config added.
Enabled OAuth only on method /record/{id}

### DIFF
--- a/src/main/openapi/ds-present-openapi_v1.yaml
+++ b/src/main/openapi/ds-present-openapi_v1.yaml
@@ -39,7 +39,7 @@ paths:
         - '${project.name}'
       summary: 'Retrieve metadata for the record with the given ID and in the given format.'
       operationId: getRecord
-      security:  #Require KeyCloak token
+      security: 
         - KBOAuth:
           - any      
       parameters:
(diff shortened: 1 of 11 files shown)


src/main/openapi/ds-present-openapi_v1.yaml
```diff
@@ -1115,18 +1118,56 @@ components:
...
+ # TODO: Inject this from properties when DRA-139 has been implemented
+  securitySchemes:
+
+    KBOAuth:
+      type: oauth2
+      description: 'This API uses OAuth 2 with the implicit grant flow. [More info](https://api.example.com/docs/auth)'
+      flows:
+        implicit:   # <---- OAuth flow(authorizationCode, implicit, password or clientCredentials)
+          authorizationUrl: 'https://keycloak-devel-01.kb.dk/realms/ds-poc/protocol/openid-connect/auth'
+          #authorizationUrl: https://api.example.com/oauth2/authorize # Overridden by the configuration
```
VictorHarbo, line 1158: Please inject from properties with syntax ${config.yaml.path}

VictorHarbo (changes requested):
Looks good. Only one comment here.

src/main/openapi/ds-present-openapi_v1.yaml
thomasegense: fixed

VictorHarbo (approved):
Looks good now

PR #169, opened 2024-07-24
Dra 727 implement ownproduction filter and refactor extractions from metadata

VictorHarbo:
### Prepare for own production filtering. 
This PR refactors how values that are used before the XSLT transformations are extracted from Preservica records. Before each value was extracted from an XML stream whenever they were needed. Now all values are extracted to a common POJO and then accessed from there. This makes the reading of the XML happen once instead of 3-5 times. **This is the cause for so many changed files**.

Besides this, the PR introduces fields for own_production in the trasnformations and in the solr schema. **NB: These are quite hard to test as we don't have the data present in the preservica stage environment yet. I've tried creating a couple of test records by hand.**

Ps. Remeber to pull newest test files from aegis.


thomasegense (approved):
Very good java doc and unittest+test data.
Business logic is beyond my knowledge.

PR #170, opened 2024-07-25
Dra 1095 add missing fields

VictorHarbo:
Adds the Preservica fields 'forfatter', 'medvirkende' and 'instruktion' to transformations and renames the image field creator_full_name to creator for better usage across collections. 

As I am drafting this text it occurs to me that 'medvirkende' in a radio show doesnt translate well to actor. contributor might be a better name in the schema.org. Would we then want to rename the solr fields for both origins? We might need to discuss this face to face.


thomasegense (approved):
LGTM! Hope the endless XLST can come to an end :)

PR #171, opened 2024-07-29
Dra 838 correct entries

VictorHarbo:
PR creating _length and _count fields for all missing text and multivalued fields respectively


thomasegense (approved):
LGTM!

PR #173, opened 2024-09-04
Dra 1115 holdback from nielsen

VictorHarbo:
Extends value extractor to handle nielsen and tvmeter content. Also lowers some logging to debug and adds rules for the holdback filtering. A review of the business logic is to be scheduled besides this code review.


thomasegense (approved):
LGTM

PR #175, opened 2024-09-10
Dra 1215 update channel names. 

VictorHarbo:
@jorntx found a bug in the channel names and after some digging I've discovered a better way to produce these values. Mappings have been discussed with Lars Lundegaard


thomasegense (approved):
Simple fix. LGTM

PR #177, opened 2024-09-17
Dra 1282 cache startup query

VictorHarbo:
The frontend application sends the following query: `?q=*&facet=true&facet.limit=-1` everytime the frontpage is loaded. I've added the query to the warming of searchers. On my local index of devel, the query time went from 300millis to 50 millis by pre-warming the query.

This query is used to build the time searching component on the frontpage of the application.


thomasegense (approved):
simple

PR #178, opened 2024-09-24
Dra 1292 origin and country of origin

VictorHarbo:
After meeting with DR it has become clear that field origin is to be used for own production calculation and field contry of origin is to be used for holdback calculation. These are not the same field as first understood. This PR adds extraction of origin and uses that value for own production calculation.


thomasegense (approved):
LGTM

PR #179, opened 2024-09-24
DRA-1296 fix transformation errors

davidgrove73:


src/test/java/dk/kb/present/TestFiles.java
```diff
@@ -59,6 +59,9 @@ public class TestFiles {
     public static final String PVICA_HOMEMADE_DOMS_MIG_WITH_TVMETER_ADDED = "internal_test_files/homemade/correctDomsMigWithTVMeterEnrichment.xml";
     public static final String PVICA_HOMEMADE_NOT_OWNPROD = "internal_test_files/homemade/badOwnProduction.xml";
     public static final String PVICA_HOMEMADE_RADIO_WITH_CONTRIBUTORS = "internal_test_files/homemade/radioWithContributors.xml";
+    public static final String PVICA_HOMEMADE_INVALID_NUMBERS = "internal_test_files/homemade/invalid_numbers.xml";
+    public static final String PVICA_HOMEMADE_VALID_NUMBERS = "internal_test_files/homemade/valid_numbers.xml";
```
VictorHarbo, line 63: Valid numbers file does not seem to be in aegis, however it isn't used in any tests. Is it supposed to be used for a test?

VictorHarbo (changes requested):
Very thorough checking of numbers in the transformations. One single comment on a missing file/test

src/test/java/dk/kb/present/TestFiles.java
davidgrove73: I wanted to make a test, to check that valid numbers still passed through the transformation. But I don't think its needed as it is covered by other tests. I have deleted the reference to the fi.e

VictorHarbo (approved):
Looks good now!

PR #182, opened 2024-09-30
Change field migrated_from to originates_from and add 'Preservica' as…

VictorHarbo:
… default value.


thomasegense (approved):
Deployed + new index build. Looks good.,

PR #185, opened 2024-10-01
Dra 1321 update solr download script

VictorHarbo:
A simpler version was used to make devel work again. This version has been tested locally it downloads versions before and after 9 as expected.


thomasegense (approved):
Looks correct. 
Will test new install maybe when increasing maxBooleanClauses....

PR #187, opened 2024-10-08
Dra 1326 mark restricted material as restricted

VictorHarbo:


thomasegense (approved):
LGTM. Better naming of fields

PR #190, opened 2024-10-16
New text_general_no_keyword field. Title uses this field now.

thomasegense:
New text_general_no_keyword field. Title uses this field now.


VictorHarbo (approved):
Looks good and works as we discussed

PR #193, opened 2024-10-21
Transformation error handling

VictorHarbo:
Update transformation error handling to include error messages from all errors.


thomasegense (commented):
Not sure I follow it all. But I understand the unittest and it seems to work.

PR #196, opened 2024-10-23
fix colon transformation error

VictorHarbo:
Branch deployed to devel and current index build from this branch


thomasegense (approved):
Working!

PR #201, opened 2024-11-05
Add field own_production to radio records

VictorHarbo:
Add own_production value to radio records. This can make license rule more simple


thomasegense (approved):
simple

PR #202, opened 2024-11-08
DRA 1438 and DRA 1453

VictorHarbo:
Add solr fields to tell if record contains metadata from added fragments

Add description as copyField in solr


thomasegense (approved):
LGTM!

PR #203, opened 2024-11-11
Dra 1428 last genre cleanup

VictorHarbo:
This PR does the following: 
* Rename genres as specified by UX Team.
* Remove genre `Unresolved` and move these values to `Rodekassen`. (1000 viewable records will be affected.)
* Add default genre `Rodekassen` for records that have no genre what so ever.

To remove the category `Sport` a ds-license restriction has to be updated. Here is the updated restriction on devel. I've added `genre:"Sport"`: 
image


thomasegense (approved):
LGTM

PR #205, opened 2024-11-20
Dra 1480 change own production value

VictorHarbo:
Currently we create the fields with the following names: 

- `own_production`
- `own_production_code`

Do you think that these should be renamed to `production_code_allowed` and `production_code_value` as including co-production and enterprise-production in a field names `own_production` could potentially be misleading? 

This would require a change in the license filter to look at `production_code_allowed` instead of `own_production`. 

What are your thoughts on this?

I've created this as a draft PR until I get your comments on the above


thomasegense (approved):
LGTM!

PR #212, opened 2024-12-16
DRA-1540

davidgrove73:
Make HoldbackDatePicker.getHoldbackDateForRecord threadsafe


VictorHarbo (approved):
Looks good. I believe you found the race condition! We should try indexing the devel env anew now.

PR #216, opened 2025-01-27
bumb version of swagger-ui

davidgrove73:
bumb swagger-ui to latest stable version to fix problems with inline view of responses


VictorHarbo (approved):
Tested on new devel machine. Behaviour is as expected now!

PR #219, opened 2025-02-03
Dra 399 stream error handling

VictorHarbo:
This PR implements two things: 
1. Implementation of an individual log file where all transformation errors are logged for easy access. 
2. Inclusion of errorList in getRecords endpoint response


thomasegense (approved):
Looks good, seems to be handled as simple as possible.

PR #220, opened 2025-02-18
Service2service OAuth2 enabling on DsPresentClient

thomasegense:
See full description here for DsStorage: 
https://github.com/kb-dk/ds-storage/pull/65

The two IIIF service methods has no java client class generation, and probably also should not have one.


VictorHarbo (approved):
Looks good

PR #223, opened 2025-03-05
Use new DsLicense and DsStorage client. Disabled broken mock unittest

thomasegense:
This module is still generating the old client used by DsDatahandler. But it is now using new DsStoage and DsLicense clients.


VictorHarbo (approved):
Looks good

PR #224, opened 2025-03-05
Dra 1718 add ds id restrictions

VictorHarbo:
Add functionality to restrict access to records based on DS IDs.


thomasegense (approved):
Very good. Just what I had expected that would required to be changed.

PR #225, opened 2025-03-07
Dra 1734 fix integration test

thomasegense:
Fix mock unittest.


VictorHarbo (approved):
Yes, Good that we found my mistake from earlier as well

PR #226, opened 2025-03-10
Dra 1746 holdback from start of next year

VictorHarbo:
PR of changing holdback calculation. DO NOT MERGE before we have a go on the change from the right place.

### DIFF
--- a/src/main/java/dk/kb/present/dr/holdback/HoldbackDatePicker.java
+++ b/src/main/java/dk/kb/present/dr/holdback/HoldbackDatePicker.java
@@ -141,11 +141,7 @@ public class HoldbackDatePicker {
                     // When holdback is more than a year, it should be calculated from the 1st of January the following year and days aren't actually needed more, so we convert
                     // days to years, to make it easier to add correct amount of time
                     int holdbackYears = holdbackDays / 365;
-
-                    ZonedDateTime cleanedStartDate = DataCleanup.getCleanZonedDateTimeFromString(startDate);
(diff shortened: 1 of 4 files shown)


src/main/java/dk/kb/present/dr/holdback/HoldbackDatePicker.java
```diff
@@ -155,6 +182,22 @@ private static String calculateHoldbackDate(ZonedDateTime startDate, int holdbac
...
+     * @return the date, when the holdback period has expired as a string in the format: yyyy-MM-dd'T'HH:mm:ssZ.
+     */
+    private static String calculateHoldbackDateByYears(ZonedDateTime startDate, int holdbackYears) {
+        ZonedDateTime holdbackExpiredDate = startDate.plusYears(holdbackYears);
+
+        // Using .ISO_INSTANT as this is solr standard
+        DateTimeFormatter formatter = DateTimeFormatter.ISO_INSTANT;
+
+        String formattedHoldbackDate = holdbackExpiredDate.format(formatter);
+        return formattedHoldbackDate;
```
kesj-kb, line 198: I think this method should be responsible for calculating the days following the algorithm and call getFirstComingJanuary() and do the calculation.
Right now the method assumes that the caller first finds the coming first january but what if the called forgot that?

Make this method take a random date as start date add the days to next coming first january and and the holdback years.

kesj-kb (changes requested):

kesj-kb (approved):

PR #227, opened 2025-03-13
Removed autogenerated client

thomasegense:
Was not so hard to fix after all. Just deleted a lot of method removed http-headers everywhere (private class attribute and all methods). It seemed to be used or making any difference. Indexing still works perfect from ds-datahandler.


src/main/java/dk/kb/present/util/DsPresentClient.java
```diff
@@ -1,178 +1,99 @@
...
-     * When working with YAML configs, it is suggested to use the structure
+     * When working with YAML configs, it is suggested to define the present URI as the structure
      * <pre>
-     * present:
-     *   url: 'http://localhost:9073/ds-present/v1'
-     *   headers:
-     *     # Placeholder for authentication until OAuth2 integration is in place
-     *     # Set this to 'internal_service` on local and devel to bypass license check in ds-present
-     *     - Simulated-OAuth2-Group: anonymous
+     * storage:
```
VictorHarbo, line 46: `storage` should be `present` here.

src/main/java/dk/kb/present/util/DsPresentClient.java
```diff
@@ -216,7 +138,8 @@ public ContinuationInputStream<Long> getRecordsJSON(String origin, Long mTime, L
             throws IOException {
         URI uri;
         try {
-            uri = new URIBuilder(serviceURI + "records")
+            uri = new URIBuilder(serviceURI)
+                    .appendPath("records")                                        
                     // .setPath overrides all paths given in the serviceURI. Should be changed using pathSegments in the future
                     //.setPath("records")
```
VictorHarbo, line 144: please remove this comment as it is not valid with the ned version of URIBuilder

VictorHarbo (changes requested):
This is definitely a much cleaner client. I wonder what the idea behind all that header manipulation was. Guess we will never know. Super small comments

src/main/java/dk/kb/present/util/DsPresentClient.java
thomasegense: done

src/main/java/dk/kb/present/util/DsPresentClient.java
thomasegense: fixed. But maybe code should be removed. Constructor no longer takes a yaml, but only the URL to the service.

VictorHarbo: Correct, there is no need to specify anthing about YAML here, good idea to re-write the javadoc all-together

VictorHarbo (changes requested):

VictorHarbo (approved):
Looks good now. Nice cleanup

PR #228, opened 2025-03-17
Oauth2 enabled in unitest

thomasegense:
Had to change aegis properties from devel11 to ds-devel01.

Remember kb init and see if unittest is also working for you.


VictorHarbo (approved):
All tests pass against the oAuth2 enabled devel machine

PR #231, opened 2025-04-04
Dra 1792 rights from license

VictorHarbo:
Implement ds-license lookup for rights in ds-present. Cleanup of old implementation comes in next PR


davidgrove73 (commented):
Looks fine except I can't se where dsIdRestricted is set :-) -

PR #234, opened 2025-04-05
Dra 1843 spellcheck bug release branch

thomasegense:


VictorHarbo (approved):
This fixes the bug. However we should look into the `SynonymGraphFilterFactory` as this filter factory seems like the one solr is continuing with in the future.

PR #239, opened 2025-05-14
Dra 1693 presentationfile from metadata

VictorHarbo:
Rememeber to pull aegis and do a kb init in ds-present to have access to the new test files


src/main/java/dk/kb/present/util/ExtractedPreservicaValues.java
```diff
@@ -160,6 +160,7 @@ public RightsCalculationInputDto asRightsCalculationInputDto(PlatformEnumDto pla
         restrictionsDto.setRecordId(getId());
         restrictionsDto.setDrProductionId(getProductionId());
         restrictionsDto.setProductionCode(getOrigin());
+        restrictionsDto.setTitle("Test title"); // TODO: Title should be sent here as well
```
kesj-kb, line 163: Test title?

src/main/resources/xslt/preservica2schemaorg.xsl
```diff
@@ -1312,56 +1333,64 @@
...
+          <xsl:value-of select="number(normalize-space(substring-after(extension , 'kanalid:')))"/>
         </xsl:variable>
-        <xsl:if test="string($channelId) != 'NaN'">
-          <f:number key="kb:channel_id">
-            <xsl:value-of select="$channelId"/>
-          </f:number>
-        </xsl:if>
+
+        <xsl:choose>
+          <xsl:when test="extensionAuthorityUsed = 'ritzau' and string($channelId) != 'NaN'">
```
kesj-kb, line 1347: Could channelId ever be an empty String or will we always have a number or "NaN"?

src/main/java/dk/kb/present/util/ExtractedPreservicaValues.java
VictorHarbo: Good catch. the title of the record should be extracted in the "preprocessing" step, before the XSLT transformation and added here, so that it can be checked by the rightsmodule. I've created the task DRA-1946 for this as that requires some more changes. Is that alright with you?

src/main/resources/xslt/preservica2schemaorg.xsl
VictorHarbo: Line 1341 checks the following: `f:string-length(normalize-space(substring-after(extension , 'kanalid:'))) > 0` which translates to _if the value after colon is not a string of lengt zero_ which I think is what you are asking for? 

As to if it is ever possible to have the empty string in hand, we normally assume that data from presevica are messy and that it could be possible to get an empty string after the colon

PR #269, opened 2026-01-21
Dra 1698 extract fuzzyfragment for rights calculation

davidgrove73:
use metadata fragment with "fuzzy"-metadata (http://id.kb.dk/schemas/supplementary_dr_tv_1973_2005_metadata) in rights calcultions. Fuzzy metadata should have lowest priority.


jfre-kb (approved):
lgtm

